PR #1, opened 2016-01-30
Add project skeleton for YARPC java

junchaowu:


breerly:
cc @yarpc/java @abhinav @kriskowal need some more eyes here

pom.xml
```diff
@@ -0,0 +1,101 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.uber.yarpc-java</groupId>
```
breerly: shouldn't this be something like `io.yarpc.java`?

pom.xml
```diff
@@ -0,0 +1,101 @@
...
+    <packaging>jar</packaging>
+    <version>0.0.1</version>
+    <name>yarpc-java</name>
+
+    <developers>
+        <developer>
+            <name>Junchao Wu</name>
+            <email>junchaow@uber.com</email>
+            <organization>Uber Technologies, Inc.</organization>
+            <organizationUrl>https://uber.github.io</organizationUrl>
```
breerly: `github.com/yarpc`*

src/main/java/example/example.java
```diff
@@ -0,0 +1,42 @@
...
+import io.yarpc.RPC;
+import io.yarpc.config.RpcConfig;
+import io.yarpc.transport.Inbound;
+import io.yarpc.transport.http.HTTPInbound;
+
+import java.util.ArrayList;
+
+public class example {
+    public static void main(String[] args) throws InterruptedException {
+        RPC rpc = new RPC(new RpcConfig(
```
breerly: suggest `yarpc.config.Config` instead of `yarpc.config.RpcConfig`

src/main/java/io/yarpc/RPC.java
```diff
@@ -0,0 +1,62 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc;
+
+
+import io.yarpc.config.RpcConfig;
```
breerly: Do we really want this project to take up the top level `io.yarpc.*` namespace? Perhaps it should be at `io.yarpc.yarpc-java.*`? @abhinav give me a second opinion.

src/main/java/io/yarpc/common/SchemeType.java
```diff
@@ -0,0 +1,57 @@
...
+ * <p/>
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.common;
```
breerly: can we do better than `common` here?

src/main/java/io/yarpc/handler/IHandler.java
```diff
@@ -0,0 +1,29 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.handler;
+
+
+import io.yarpc.Context;
+
+public interface IHandler<I, O> {
```
breerly: Prepending the interface `I`? Seems not necessary; is this even still a thing in J-land?

src/main/java/io/yarpc/model/InternalRequest.java
```diff
@@ -0,0 +1,79 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.model;
+
+
+public class InternalRequest {
```
breerly: `s/InternalRequest/Request`

src/main/java/io/yarpc/model/InternalResponse.java
```diff
@@ -0,0 +1,39 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.model;
+
+
+public class InternalResponse {
```
breerly: `s/InternalResponse/Response`

breerly:
Take care of the naming things, then :ship:

breerly:
Please get at least 2 ships

src/main/java/example/example.java
junchaowu: In this case I would move the Config.java out of config package. just put it in yarpc root package directory.

src/main/java/io/yarpc/handler/IHandler.java
junchaowu: This is a convention I follow all the time in Java.

src/main/java/io/yarpc/model/InternalRequest.java
junchaowu: This is class I move from previous code. I will rename or relocate it in a better way when I finally use it in this new project.

src/main/java/io/yarpc/handler/IHandler.java
breerly: I don't believe it's an accepted convention. @abhinav whatchuthink?

http://stackoverflow.com/questions/2814805/java-interfaces-implementation-naming-convention

src/main/java/io/yarpc/transport/http/HttpServerInboundHandler.java
```diff
@@ -0,0 +1,19 @@
...
+public class HTTPServerInboundHandler extends SimpleChannelInboundHandler<HttpObject> {
+
+    @Override
+    protected void messageReceived(ChannelHandlerContext channelHandlerContext, HttpObject httpObject) throws Exception {
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) { // (4)
+        // Close the connection when an exception is raised.
+        cause.printStackTrace();
```
anson627: we'd better log this once we have a logger in place.

src/main/java/io/yarpc/HandlerRegistry.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package io.yarpc;
+
+
+import io.yarpc.handler.IHandler;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class HandlerRegistry implements IHandlerRegistry {
```
abhinav: As discussed, please rename the interface to HandlerRegistry and the class can be MapHandlerRegistry.

src/main/java/io/yarpc/handler/IHandler.java
anson627: I'm not a big fan of interface either. I usually make an interface until I really need one. (more than 3 derived classes)

anson627:
accepted with minor comments

src/main/java/io/yarpc/MapHandlerRegistry.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final ConcurrentMap<String, Handler> handlers = new ConcurrentHashMap<String, Handler>();
+
+
+    public void addHandler(String procedure, Handler handler) {
+        handlers.put(procedure, handler);
+    }
+
+    public void removeHandler(String procedure) {
+        handlers.remove(procedure);
```
abhinav: Since it's early enough in the project, I'm gonna suggest enforcing some best practices from the start. Here's one:

Use [`Objects.requireNotNull`](https://docs.oracle.com/javase/7/docs/api/java/util/Objects.html#requireNonNull(T,%20java.lang.String)) fields that can't be null when entering methods. Helps fail fast and find the exact place a null value was introduced.

Alternatively/additionally, add `javax.annotation.Nullable` and/or `javax.annotation.Nonnull` annotations where appropriate to at least help static analyzers like findbugs. (Pick one strategy and use it consistently.)

src/main/java/io/yarpc/MapHandlerRegistry.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public Handler lookupHandler(String procedure) {
+        if (procedure == null) {
+            return null;
+        }
+        return handlers.get(procedure);
```
abhinav: Objects.requireNotNull(procedure, "procedure is required");

src/main/java/io/yarpc/RPC.java
```diff
@@ -0,0 +1,62 @@
...
+import io.yarpc.handler.Handler;
+import io.yarpc.model.RequestBody;
+import io.yarpc.model.ResponseBody;
+import io.yarpc.transport.Inbound;
+
+import java.util.List;
+
+public class RPC {
+    private Config config;
+    private MapHandlerRegistry mMapHandlerRegistry;
```
abhinav: No, this only depends on `HandlerRegistry`. There's no map-specific APIs being used here. In the constructor, you can build and assign a MapHandlerRegistry to it by default but make the class itself only use the interface.

src/main/java/io/yarpc/exception/RpcException.java
```diff
@@ -0,0 +1,28 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.exception;
+
+public class RpcException extends Exception {
+    public RpcException(String message) {
+        super(message);
+    }
```
abhinav: If you don't do anything here, it'll have this constructor anyway.

abhinav: Plus it'll also have the ones where you can specify the underlying exception as the second argument.

src/main/java/io/yarpc/handler/Handler.java
```diff
@@ -0,0 +1,29 @@
...
+ * THE SOFTWARE.
+ */
+package io.yarpc.handler;
+
+
+import io.yarpc.Context;
+
+public interface Handler<I, O> {
+    I handle(Context context, O request);
+}
```
abhinav: Is the order of I/O inverted? Sounds like I should be input and O output but this is doing the opposite.

src/main/java/io/yarpc/model/RequestBody.java
```diff
@@ -0,0 +1,27 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.model;
+
+
+public abstract class RequestBody<T> {
+    public abstract T getBody();
+}
```
abhinav: Why not interface? Only method here is abstract.

src/main/java/io/yarpc/model/ResponseBody.java
```diff
@@ -0,0 +1,27 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.model;
+
+
+public abstract class ResponseBody<T> {
+    public abstract T getBody();
+}
```
abhinav: Same question

src/main/java/io/yarpc/transport/http/HttpServerInboundHandler.java
abhinav: +1.

I recommend [logback](http://logback.qos.ch/).

PR #3, opened 2016-02-03
Add JSON inbound handler support

junchaowu:


src/main/java/example/http/example.java
```diff
@@ -19,17 +19,20 @@
...
 
 
+import example.http.json.UserHandler;
 import io.yarpc.RPC;
 import io.yarpc.config.Config;
+import io.yarpc.encoding.json.JSONInboundHandlerWrapper;
 import io.yarpc.transport.Inbound;
 import io.yarpc.transport.http.HTTPInbound;
 
 import java.util.ArrayList;
```
abhinav: I don't think you want the example to part of your library source code.

You want to put in a separate top-level example directory and not have it shipped with your library.

src/main/java/io/yarpc/HandlerRegistry.java
```diff
@@ -22,12 +22,12 @@
...
+import io.yarpc.handler.HandlerWrapper;
 
 public interface HandlerRegistry {
-    void addHandler(String procedure, Handler handler);
+    void addHandler(String procedure, HandlerWrapper handler);
 
     void removeHandler(String procedure);
 
-    Handler lookupHandler(String procedure);
+    HandlerWrapper lookupHandler(String procedure);
```
abhinav: I haven't gotten to the relevant code yet but from the name alone it sounds like `HandlerWrapper` shouldn't be part of a public interface.

src/main/java/io/yarpc/context/http/ContentType.java
```diff
@@ -0,0 +1,75 @@
...
+package io.yarpc.context.http;
+
+
+import io.yarpc.encoding.EncodingType;
+
+public enum ContentType {
+    RAW("application/octet-stream"),
+    JSON("application/json"),
+    THRIFT("application/thrift");
+
```
abhinav: How will encodings not built into the library provide the content type? Sounds like this should be a transport property that the scheme provides?

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
```diff
@@ -0,0 +1,108 @@
...
+    public ObjectMapper getMapper() {
+        return mapper;
+    }
+
+    public Handler getHandler() {
+        return handler;
+    }
+
+    public Class<?> getReqT() {
+        return reqT;
```
abhinav: Styling: Use full names. `getRequestType`/`getRequestClass` and `getResponseType`/`getResponseClass`

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public Class<?> getResT() {
+        return resT;
+    }
+
+    private Class<?> resT;
```
abhinav: Please have consistent ordering in your class declarations.

    <attributes>

    <methods>

    <attributes>

<constructor>

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public JSONInboundHandlerWrapper(Handler handler) {
+        parseJSONHandler(handler);
+        mapper = new ObjectMapper();
```
abhinav: Calling instance methods from constructors is generally frowned upon.

Consider setting up findbugs to run over the code as part of your Travis build.

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private void parseJSONHandler(Handler handler) {
+        TypeResolver typeResolver = new TypeResolver();
+        List<ResolvedType> types = typeResolver.resolve(handler.getClass()).findSupertype(Handler.class)
+                .getTypeParameters();
+        reqT = types.get(0).getErasedType();
+        resT = types.get(1).getErasedType();
+        this.handler = handler;
+    }
```
abhinav: Why is this an instance method at all. The functionality itself should be in a static method somewhere and the results (`reqT` and `resT`) should be returned.

Don't initialize the current object in methods. Instance methods should only be called after the object has been instantiated.

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
abhinav: Styling: Always use `this` when assigning to instance variables. Helps with readability and avoids bugs from accidental shadowing of instance variables.

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public <T> T unmarshal(String str, Class<? extends T> clazz) throws IOException {
+        return mapper.readValue(str, clazz);
+    }
+
+    @Override
+    public EncodingType getEncodingType() {
+        return EncodingType.JSON;
+    }
```
abhinav: What value is the EncodingType serving? By adding an enum you're making it so that new encodings require changing the library and can't be separate.

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
```diff
@@ -0,0 +1,108 @@
...
+                    .headers(response.getHeaders()).build();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+        return null;
+    }
+
+    private void validate(Context context) throws RPCException {
+        if (context.getEncodingType() != getEncodingType()) {
+            throw new RPCException(String.format("Content type mismatch registered handle's type."));
```
abhinav: IMO client-side validation errors should use IllegalArgumentException .

src/main/java/example/http/example.java
anson627: there is no way to ask maven to not package your example code into a jar, you can put it into a test folder if you really don't want to ship your example. In Java, usually people ship examples along with jar, so consumers can run you jar with example class name to learn how to use it.

src/main/java/io/yarpc/handler/HandlerWrapper.java
```diff
@@ -0,0 +1,29 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.handler;
+
+
+import io.yarpc.encoding.EncodingType;
+
+public interface HandlerWrapper extends Handler<String, String> {
+    EncodingType getEncodingType();
```
abhinav: What's this class supposed to do/why does it exist?

src/main/java/io/yarpc/model/Response.java
```diff
@@ -0,0 +1,74 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.model;
+
+import java.util.Map;
+
+public class Response<BODYT> {
```
abhinav: Not a fan of this all-caps generic argument naming convention.

src/main/java/io/yarpc/model/Response.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public BODYT getBody() {
+        return body;
+    }
+
+    public void setBody(BODYT body) {
+        this.body = body;
+    }
+
+    public static class ResponseBuilder<BODYT> {
```
abhinav: I would make this a class in its own file. If you're going to leave it in this class, it would make sense to call it just `Builder` so you can do `new Response.Builder<..>()`.

src/main/java/io/yarpc/model/Response.java
```diff
@@ -0,0 +1,74 @@
...
+            this.nestedRespBody = body;
+            return this;
+        }
+
+        public ResponseBuilder headers(Map<String, String> headers) {
+            this.nestedHeaders = headers;
+            return this;
+        }
+
+        public Response build() {
```
abhinav: Where's the generic type for Response? You need to include that in the result.

Do you have warnings enabled in your IDE? It would warn you about this. Don't ignore those warnings. They're all valid. Fix them.

src/main/java/example/http/example.java
abhinav: Oh sorry I wasn't clear. I'm happy to ship example code in the JAR. I meant as part of the packages that get installed. Right now, if I depend on this code, I'll get an unexpected `example.http` package.

src/test/java/io/yarpc/encoding/json/TestBean.java
```diff
@@ -0,0 +1,48 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.encoding.json;
+
+public class TestBean {
```
anson627: can we consider using mockito? so you don't have to hand coded a lot these test/mock classes. totally optional.

src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            handlerWrapper) {
+        Response<String> response = null;
+        HttpHeaders httpHeaders = new DefaultHttpHeaders();
+        // TODO build yarpc common context including headers, tracing, meta info.
+        Context context = ContextBuilder.build(httpRequest);
+        try {
+            response = handlerWrapper.handle(context, httpRequest.content().toString(CharsetUtil.UTF_8));
+        } catch (RPCException e) {
+            e.printStackTrace();
+        }
```
abhinav: Until you have logging and error handling set up, rather than printing and moving on, just rethrow the exceptions after wrapping them in a RuntimeException (so that you don't have to have a checked exception).

src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -0,0 +1,111 @@
...
+
+
+        if (response != null) {
```
abhinav: reduce nesting here.

    if (response == null) {
        return
    }

    FullHttpResponse ...

abhinav: Although I don't think there's any valid case where `response` would be null. The previous block of code should simply fail if there's no response.

src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            FullHttpResponse httpResponse = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK,
+                    Unpooled.wrappedBuffer(response.getBody().getBytes(CharsetUtil.UTF_8)));
+            httpHeaders.set(
+                    HttpHeaderNames.CONTENT_TYPE, ContentType.fromEncodingType(context.getEncodingType()).getType());
```
abhinav: With this many `.` and `()`s, consider using local variables

src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -0,0 +1,111 @@
...
+            httpHeaders.set(HttpHeaderNames.CONTENT_LENGTH, String.valueOf(httpResponse.content().readableBytes()));
+            for (String key : response.getHeaders().keySet()) {
+                httpHeaders.set(key, response.getHeaders().get(key));
+            }
```
abhinav: How are you handling missing keys?

src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -0,0 +1,111 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.transport.http;
+
+import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.SimpleChannelInboundHandler;
+import io.netty.handler.codec.http.*;
```
abhinav: Don't import `*`. Configure your IDE to use explicit imports of everything.

src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -0,0 +1,111 @@
...
+        if (handlerWrapper.getEncodingType() != encodingType) {
+            throw new RPCException(String.format("Content type mismatch registered handle's type."));
+        }
+        return handlerWrapper;
+    }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) { // (4)
+        // Close the connection when an exception is raised.
+        cause.printStackTrace();
```
abhinav: again, wrap and rethrow. much better than print and move on.

Guava provides  `Throwables.propagate()` for this. If you want to avoid Guava as a dependency, just create your own version of that somewhere in the library and use that.

src/test/java/io/yarpc/encoding/json/TestBean.java
abhinav: +1

An using mockito/EasyMock for testing interfaces

Although I imagine you'd still have to write basic classes like this one by hand.

FWIW if you were using lombok this class would just be

    @Data public class TestBean { private String a; private int b; }

src/main/java/io/yarpc/context/http/ContentType.java
junchaowu: The Content Type is coming from our context propagation docs. That's why I put it in the context.http package.  I don't know what's your question.

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
junchaowu: Not really work on wrap as specific exception yet. Everything I will call RPCException for now.

src/main/java/io/yarpc/handler/HandlerWrapper.java
junchaowu: This will provide a consistent wrap for handlers.
The transport layer will get raw String body from http request and pass it to handlerwrapper (here we take JSON for example). The JSON handler wrapper will takes the string body, parse it as json to user's procedurer. It gets a Response from user with Json body, serialize it into String and return a Response with String body to the transport layer. 

The HandlerWrapper, thus, will provide a consistent interface layer, even if user would like to create their own handlerwrapper (a new encoding type). As long as follow the interface by taking a string body and returns a string body. The inbound framework will support that.

src/main/java/io/yarpc/context/http/ContentType.java
breerly: We don't want the transport to have knowledge of every encoding.

Let's hold off on setting the http `Content-Type` header for now until we decide that we need to do it in all langs.

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
breerly: +1 here - `Request` and `Response` are top-level objects, I'd rather them not be abbreviated anywhere.

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
breerly: +1

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
breerly: +1 on findbugs & checkstyle
+1 on using `this`
--

Didn't realize calling instance methods during construction was a bad practice - in fact, I thought it was common to delegate in a constructor.

abhinav: There are many scenarios when it's okay to call methods to help initialize an object but generally it should be avoided. It's easy for someone to come into that codebase and forget that they don't have a fully initialized object, call 5 other methods (whether on the same class or others) and end up breaking the universe.

breerly: I can buy that :+1:

.travis.yml
```diff
@@ -0,0 +1 @@
+language: java
```
breerly: Add the badge to your README as well.

anson627:
accepted with minor comments

src/main/java/example/http/example.java
```diff
@@ -37,6 +40,10 @@ public static void main(String[] args) throws InterruptedException {
                     add(new HTTPInbound());
                 }}));
 
+        rpc.register("hello", new JSONInboundHandlerWrapper(new u1()));
```
breerly: Not wild about using Wrapper here

breerly: The naming, rather.

src/main/java/example/http/json/User.java
```diff
@@ -0,0 +1,58 @@
...
+    public String getAge() {
+        return age;
+    }
+
+    public void setAge(String age) {
+        this.age = age;
+    }
+
+    public String getSex() {
+        return sex;
```
breerly: Let's go with just `age` and `name` for now - makes for a more succinct example.

src/main/java/io/yarpc/encoding/json/JSONInboundHandlerWrapper.java
junchaowu: Will setup findbugs & checkstyle in following diffs.

breerly:
I really want to see the top-level `Request` and `Response` objects taken out of the `model` package to match the other languages.

src/main/java/io/yarpc/Response.java
```diff
@@ -0,0 +1,74 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc;
+
+import java.util.Map;
+
+public class Response<BodyT> {
```
breerly: thx for moving this here! :100:

breerly:
This is looking like a pretty good starting point!

:sheep: :it:

PR #12, opened 2016-02-11
Separate the project into two projects: yarpc-core and yarpc-example

junchaowu:
Separate the projects into yarpc-core and yarpc-example. So the yarpc-example can use the jar built from yarpc-core to show all kinds of client and server examples.

Please ignore the build at this moment because more changes will come through the following patches such as thrift integration, top request/response and so on.


junchaowu:
cc @yarpc/yarpc

breerly:
I'm not opposed to this. @abhinav you onboard?

Get tests to pass and I'm :sheep:

junchaowu:
Add thrift idl into example project to let the build pass.

kriskowal:
:ok:

kriskowal:
To be clear, I am not in love with this approach. I feel like the root of the repo should be the library project and that the example should be in a subdirectory, albeit a separate project. But as long as we’re flexible, this is fine.

breerly:
I was hoping to have @abhinav to look at this before merging.

build/travis/install-thrift.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -e
+wget http://archive.apache.org/dist/thrift/0.9.2/thrift-0.9.2.tar.gz
+tar -xzvf thrift-0.9.2.tar.gz
+cd thrift-0.9.2
+./configure --with-cpp=no --with-c_glib=no --with-java=yes --with-ruby=no --with-erlang=no --with-go=no --with-nodejs=no
```
abhinav: Why do you have this if you're already asking for thrift-compiler to be installed by apt in your travis.yml?

abhinav:
k

junchaowu:
@kriskowal this approach is quite common for maven project. Use the top level maven to maintain couple sub module maven projects. To separate the example with core library, the example can use the API through the library but not through directly accessing source code.

build/travis/install-thrift.sh
junchaowu: I am not sure about that part. It is used in tchannel-java. In next diff, I will try to remove one of them to see if build passes.

PR #14, opened 2016-02-11
Json encoding for http client & server.

junchaowu:
Added Top level request and response (as we discussed for four language)

Added Transport level request and response. (understand current circumstance that the context and non-context are not quite well defined , I just use Map<String, Object> to store all kinds of attributes for request and response in transportation layer.)  objects like class, string, array or lists, in java can be converted to Object type, when it's being used converted back to the necessary type specifically.

Added Http outbound for Json request.


yarpc-core/src/main/java/io/yarpc/RPCFuture.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * RPCFuture builds the bridge between the caller thread sync operations & async operations.
+ * @param <V> Type of value saved in the future.
+ */
+public final class RPCFuture<V> extends AbstractFuture<V> {
+
+    private static final Logger logger = LoggerFactory.getLogger(RPCFuture.class);
+
+    public RPCFuture() {
+    }
```
abhinav: Empty constructor is not necessary if you have no constructors.

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
```diff
@@ -0,0 +1,55 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.ExecutionException;
+
+/**
+ * RPCFuture builds the bridge between the caller thread sync operations & async operations.
+ * @param <V> Type of value saved in the future.
+ */
+public final class RPCFuture<V> extends AbstractFuture<V> {
```
abhinav: How is this class different from [SettableFuture](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/util/concurrent/SettableFuture.html)? Why have our own?

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
```diff
@@ -0,0 +1,55 @@
...
+        return super.set(value);
+    }
+
+    @Override
+    public boolean setException(Throwable throwable) {
+        return super.setException(throwable);
+    }
+
+    public void sync() throws ExecutionException, InterruptedException {
+        get();
```
abhinav: `get()` already blocks if necessary and returns the result. This method is not required.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -23,39 +23,62 @@
...
-
-    public void start(int port) {
+    public RPCFuture<Void> start() {
+        this.rpcFuture = new RPCFuture<Void>();
         List<Inbound> inbounds = this.config.getInbounds();
         for (Inbound inbound : inbounds) {
             try {
                 inbound.setHandlerRegistry(this.mHandlerRegistry);
-                inbound.start(port);
+                futures.add(inbound.start());
```
abhinav: Please always specify `this.` when accessing/changing instance variables. You can make your IDE do that for you.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -23,39 +23,62 @@
...
     public void register(String proceduer, HandlerWrapper handler) {
         this.mHandlerRegistry.addHandler(proceduer, handler);
     }
 
-    public void start() {
-        start(this.config.getPort());
-    }
-
-    public void start(int port) {
+    public RPCFuture<Void> start() {
```
abhinav: Document that a future is returned which resolves when the RPC is stopped.

I understand this is early code and you don't need to document everything at this stage, but information like this, which may be needed by the reviewer should be included.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -23,39 +23,62 @@
...
+import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.Future;
 
 public class RPC {
+    private final Logger logger = LoggerFactory.getLogger(RPC.class);
     private Config config;
     private HandlerRegistry mHandlerRegistry;
+    private List<Future> futures;
+    private RPCFuture rpcFuture;
```
abhinav: Is the idea for this being an instance variable that you will simply return the already running future if `start()` is called again? If yes, do that.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -23,39 +23,62 @@
...
 
+import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.Future;
 
 public class RPC {
+    private final Logger logger = LoggerFactory.getLogger(RPC.class);
     private Config config;
     private HandlerRegistry mHandlerRegistry;
+    private List<Future> futures;
```
abhinav: Why store this at all? Generally I recommend going for a store-only-if-required approach. This can be local to `start()`.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -23,39 +23,62 @@
...
+import io.yarpc.transport.Outbound;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.Future;
 
 public class RPC {
+    private final Logger logger = LoggerFactory.getLogger(RPC.class);
```
abhinav: Forgot a `static` there

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -23,39 +23,62 @@
...
         this.mHandlerRegistry.addHandler(proceduer, handler);
     }
 
-    public void start() {
-        start(this.config.getPort());
-    }
-
-    public void start(int port) {
+    public RPCFuture<Void> start() {
+        this.rpcFuture = new RPCFuture<Void>();
```
abhinav: We can make this idempotent. Check if `rpcFuture` is already set and running, and if so, return it.

yarpc-core/src/main/java/io/yarpc/RPC.java
abhinav: Choose a better name for this. `stopFuture` or `runningFuture`. `rpcFuture` is generic given that the type itself is `RPCFuture`.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -23,39 +23,62 @@
...
             } catch (Exception e) {
-                e.printStackTrace();
+                this.rpcFuture.setException(e);
             }
         }
+        return this.rpcFuture;
     }
 
+    public Outbound lookUpOutbound(String serviceName) {
+        return this.config.getOutbounds().get(serviceName);
```
abhinav: Can this return `null`? Is that valid? Throw an exception if not.

junchaowu: This will return ``null``. The caller should tell the return value. The check will be on caller side.

yarpc-core/src/main/java/io/yarpc/RPCClient.java
```diff
@@ -0,0 +1,28 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc;
+
+
+public abstract class RPCClient {
+    public abstract <ReqBodyT, RespBodyT> RPCFuture<Response<RespBodyT>> send(
+            Request<ReqBodyT> request, Class<RespBodyT> responseBodyType) throws Exception;
```
abhinav: Something we discussed in the other PR. `ReqBodyT` and `RespBodyT` are reassy noisy. Use simple one letter variable names. `I` and `O` works well here.

    public abstract <I, O> RPCFuture<Response<O>> send(
            Request<I> request, Class<O> responseBodyType) throws Exception;

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
junchaowu: sync is another common method in Future in java. For instance TChannel-Java's future provides sync method. The Netty's Future also provide Future.sync method. It just convenient for people who are used to the sync api.

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
```diff
@@ -0,0 +1,55 @@
...
+
+
+import com.google.common.util.concurrent.AbstractFuture;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.ExecutionException;
+
+/**
+ * RPCFuture builds the bridge between the caller thread sync operations & async operations.
```
abhinav: Nit: Set your IDE to 80 character wide docstrings. Helps with readability.

Keep code lines 120 characters long (since this is Java).

yarpc-core/src/main/java/io/yarpc/Request.java
```diff
@@ -0,0 +1,163 @@
...
+ */
+package io.yarpc;
+
+import io.yarpc.context.Context;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class Request<BodyT> {
```
abhinav: Rename `BodyT` to `T`

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
junchaowu: They almost same right. I want to have RPCFuture in case we want to have custom logic in the future operations.

yarpc-core/src/main/java/io/yarpc/RPC.java
abhinav: This contract needs to be explicitly documented. Also, I would prefer that check to be on the `lookupOutbound` side. Since these lookups are happening at service startup when the clients are being set up, I would much rather have a more meaningful exception thrown by us than a NullPointerException because the caller didn't check the outbound before using it.

(Side: Go will just panic if service is not known.)

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
abhinav: `sync` is not a method on [Java Futures](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html) because they already provide a way to block for the result when you need synchronous semantics: `get()`

[Netty Futures](http://netty.io/4.0/api/io/netty/util/concurrent/Future.html) need this method because they don't have any other method to block until this future is done.

This future is a Guava `ListenableFuture` which inherits Java's `Future` type, so it has a `get()` to block for the result. Therefore, the `sync()` method is unnecessary.

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
abhinav: I would rather not yet another future type that people have to deal with. Let's use either Java's `Future`, Guava's `ListenableFuture`, or Netty's `Future`. It's easy to write adapters for the standard Future types.

yarpc-core/src/main/java/io/yarpc/Request.java
```diff
@@ -0,0 +1,163 @@
...
+    private Context context;
+    private String id;
+    private Map<String, String> headers;
+    private String encodingType;
+    private String serivce;
+    private String procedure;
+    private int port;
+    private String ip;
+
+    public Request(Builder<BodyT> builder) {
```
abhinav: Better approach: Make the constructor private and make the Builder class call the private constructor on `build()`.

    private Request(...) { .. }

    public static class Builder<T> {
        public Request<T> build() { .. }
    }

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
junchaowu: Abstractfuture extends from ListenableFuture. Even the tchannel-java implements its own future from AbstractFuture.

junchaowu: All of they are extends from Top level interface java.util.concurrent.Future. So I wouldn't worry about compatibility issue.

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
junchaowu: That's exact why we need ``sync``. ``get`` will wait and you want to get the result from the Future. ``sync`` I added here will also wait but for people who just want to sync but get things from Future. I am using Netty here. I would like to provide it. People still can use the ``get``. It is convenient for people who are used to Netty's future sync.

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: good point.

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: hmm. I create a specific exception for it.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -23,14 +23,20 @@
...
 import io.yarpc.handler.HandlerWrapper;
 import io.yarpc.transport.Inbound;
+import io.yarpc.transport.Outbound;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.List;
 
 public class RPC {
+    private static final Logger logger = LoggerFactory.getLogger(RPC.class);
```
breerly: consider making this injectable in a future diff

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -41,21 +47,51 @@ public void register(String proceduer, HandlerWrapper handler) {
...
-                inbound.start(port);
             } catch (Exception e) {
-                e.printStackTrace();
+                this.runningFuture.setException(e);
             }
         }
+        return this.runningFuture;
+    }
+
+    public Outbound lookUpOutbound(String serviceName) throws OutboundNotFoundException {
```
breerly: `s/serviceName/service/`

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -41,21 +47,51 @@ public void register(String proceduer, HandlerWrapper handler) {
...
+            throw new OutboundNotFoundException(String.format("Cannot find outbound by service name: %s", serviceName));
+        }
+        return outbound;
     }
 
+    /**
+     * Stop all the registered inbound servers.
+     * Set the future get from {@link #start()} to be done.
+     */
+    public void stop() {
```
breerly: oh, nice :+1:

breerly: still have to do this in yarpc-python

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -41,21 +47,51 @@ public void register(String proceduer, HandlerWrapper handler) {
...
+        if (this.runningFuture == null || this.runningFuture.isCancelled() || this.runningFuture.isDone())
+            return;
 
+        List<Inbound> inbounds = this.config.getInbounds();
```
breerly: Don't use the config passed the constructor - pull the pieces off this object needs and set them explicitly. 

In other words, you want the loosest possible coupling to the config object as possible.

yarpc-core/src/main/java/io/yarpc/RPCClient.java
```diff
@@ -0,0 +1,28 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc;
+
+
+public abstract class RPCClient {
+    public abstract <I, O> RPCFuture<Response<O>> send(
```
breerly: Any way to just call this a `Future` instead of `RPCFuture`?

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
breerly: If you wanted `sync` without a return value, wouldn't you just call `get` and not store the value?

yarpc-core/src/main/java/io/yarpc/Request.java
```diff
@@ -0,0 +1,163 @@
...
+ */
+package io.yarpc;
+
+import io.yarpc.context.Context;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class Request<T> {
```
breerly: Really digging these top-level objects :+1: 

Consider a library that does the getters and setters for you (until we have custom behavior there).

yarpc-core/src/main/java/io/yarpc/Response.java
```diff
@@ -21,19 +21,23 @@
...
 
+import java.util.HashMap;
 import java.util.Map;
 
-public class Response<BodyT> {
+public class Response<T> {
     // temporary use map to save headers
     private Map<String, String> headers;
-    private BodyT body;
+    private T body;
```
breerly: What is `T`? Let's come up with a more descriptive name than that if possible - I can help brainstorm.

yarpc-core/src/main/java/io/yarpc/context/Names.java
```diff
@@ -0,0 +1,36 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.context;
+
+/**
+ * Names contains a list of attributes names that will be used in the RPC transport layer.
+ */
+public class Names {
```
breerly: thanks, these make sense to consolidate :+1:

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -0,0 +1,80 @@
...
+    public static JsonClient newInstance(RPC rpc) {
+        return new JsonClient(rpc);
+    }
+
+    @Override
+    public <I, O> RPCFuture<Response<O>> send(Request<I> request, final Class<O>
+            responseBodyType) throws Exception {
+        Outbound outbound = this.rpc.lookUpOutbound(request.getSerivce());
+        request.setEncodingType(EncodingType.JSON);
+        TRequest tRequest = Transformer.toTRequest(request, JsonClient.serializer);
```
breerly: A bit confused by all these t's everywhere...

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonSerializer.java
```diff
@@ -0,0 +1,41 @@
...
+package io.yarpc.encoding.json;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.netty.util.CharsetUtil;
+import io.yarpc.encoding.Serializer;
+
+import java.io.IOException;
+
+public class JsonSerializer implements Serializer {
```
breerly: Nice, having a single serializer that is used in the inbound and outbound is :+1:

yarpc-core/src/main/java/io/yarpc/transport/TRequest.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package io.yarpc.transport;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Request representation in transport layer.
+ */
+public class TRequest {
```
breerly: So.. this is a transport-level request... why not use the top-level request directly?

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: This piece of code is still temporary . I haven't put much time on it yet. Will change config in the future diff.

yarpc-core/src/main/java/io/yarpc/RPCClient.java
junchaowu: This is try to avoid the name conflict. Future is likely meant to java.util.concurrent.Future interface. Give yarpc a specific future name will avoid any confusion.

breerly: I see...

Does the package not provide enough namespacing?

yarpc-core/src/main/java/io/yarpc/Request.java
junchaowu: I have created this task: https://github.com/yarpc/yarpc-java/issues/15 .

yarpc-core/src/main/java/io/yarpc/Response.java
junchaowu: For generics , it is totally fine to use short name. @abhinav  your thoughts

yarpc-core/src/main/java/io/yarpc/Request.java
breerly: thx :)

breerly:
I mostly had nits - if you're willing to create a task for decoupling the config from the RPC object, then I'd be a :shipit:

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: Task has been created  https://github.com/yarpc/yarpc-java/issues/16

yarpc-core/src/main/java/io/yarpc/RPCClient.java
junchaowu: Event the Yarpc, it ultimately inherits from the ``java.utl.concurrent.Future``

yarpc-core/src/main/java/io/yarpc/Response.java
abhinav: @breerly IMO short names make sense if there's a single argument or even two with obvious meanings. If you have more than that with more complex behaviors attached to them, I see the point of longer names. But in this case, I much prefer `T` over `BodyT`. `Body` would be less disagreeable but `T` is still simpler. You're supposed to document the generic type names in your javadoc anyway.

yarpc-core/src/main/java/io/yarpc/RPC.java
abhinav: Do you really need to inject Logger instances? You can configure them really extensively with your logger config files.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -41,21 +47,51 @@ public void register(String proceduer, HandlerWrapper handler) {
...
+    public RPCFuture<Void> start() {
+        this.runningFuture = new RPCFuture<Void>();
         List<Inbound> inbounds = this.config.getInbounds();
         for (Inbound inbound : inbounds) {
             try {
                 inbound.setHandlerRegistry(this.mHandlerRegistry);
-                inbound.start(port);
             } catch (Exception e) {
-                e.printStackTrace();
+                this.runningFuture.setException(e);
```
abhinav: setHandlerRegistery doesn't throw any exceptions. You probably meant to remove this?

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -41,21 +47,51 @@ public void register(String proceduer, HandlerWrapper handler) {
...
+        Outbound outbound = this.config.getOutbounds().get(service);
+        if (outbound == null) {
+            throw new OutboundNotFoundException(String.format("Cannot find outbound by service name: %s", service));
+        }
+        return outbound;
     }
 
+    /**
+     * Stop all the registered inbound servers.
+     * Set the future get from {@link #start()} to be done.
```
abhinav: For the second line:

> The future returned by `start()` will resolve to `null`.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -41,21 +47,51 @@ public void register(String proceduer, HandlerWrapper handler) {
...
+     */
+    public void stop() {
+        if (this.runningFuture == null || this.runningFuture.isCancelled() || this.runningFuture.isDone())
+            return;
```
abhinav: Always add `{` and `}` for all blocks even if they contain only one statement. It's easy to add a statement below it and forget to add the braces. You can get your IDE to always use braces for all blocks in the code style settings.

yarpc-core/src/main/java/io/yarpc/Request.java
abhinav: FWIW lombok will also do the builder for you with [`@Builder`](https://projectlombok.org/features/Builder.html).

This class will boil down to:

```
@Builder
@Data
public class Request<T> {
    private T body;
    private Context context;
    private String id;
    private Map<String, String> headers;
    private String encodingType;
    private String serivce;
    private String procedure;
    private int port;
    private String ip;
}
```

yarpc-core/src/main/java/io/yarpc/Response.java
```diff
@@ -21,19 +21,23 @@
...
-        this.headers = headers;
-        this.body = body;
+    private Response(Builder<T> builder) {
+        this.headers = builder.nestedHeaders;
+        this.body = builder.nestedRespBody;
     }
 
     public Map<String, String> getHeaders() {
+        if (headers == null) {
+            headers = new HashMap<>();
```
abhinav: I'm not a fan of the getter mutating an instance attribute here. Do this on construction and if necessary, on setHeaders. Though, better yet, just handle the case where `headers` is null gracefully in the callers.

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: I agree. The whole purpose to use slf4j library for logging is because it is easy to config to use any logger methods.

yarpc-core/src/main/java/io/yarpc/encoding/Serializer.java
```diff
@@ -0,0 +1,28 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.encoding;
+
+public interface Serializer {
```
abhinav: How generic is this? Is this used for JSON only? Will you be able to use this for Thrift?

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -0,0 +1,80 @@
...
+public class JsonClient extends RPCClient {
+    private static Serializer serializer = new JsonSerializer();
+    private final RPC rpc;
+
+    private JsonClient(RPC rpc) {
+        this.rpc = rpc;
+    }
+
+    public static JsonClient newInstance(RPC rpc) {
+        return new JsonClient(rpc);
```
abhinav: Didn't you want to depend on an actual Outbound here? The lookup shouldn't happen per-call.

abhinav: Also, `newInstance` is not the best name for a static constructor, especially if the static constructor isn't doing anything special. I would just remove the static constructor and use the JsonClient constructor.

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: Added the config in IDE.

yarpc-core/src/main/java/io/yarpc/Response.java
junchaowu: I am going to resolve this in the diff using lombok.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -0,0 +1,80 @@
...
+        outbound.send(tRequest, new ResponseListener() {
+            @Override
+            public void onResponseReceived(TResponse tResponse) {
+
+                try {
+                    O respBody = JsonClient.serializer.unmarshal(
+                            (byte[]) tResponse.get(Names.BODY), responseBodyType);
+                    Response response = new Response.Builder<O>()
+                            .body(respBody)
+                            .build();
```
abhinav: Consider using an explicit private static internal class (which can later be moved to a package-local class if necessary). Making it static makes sure that there's an explicit list of dependencies. Anonymous classes are okay if the behavior is really small but this seems like it's involved enough to deserve a named class.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -0,0 +1,80 @@
...
+                    rpcFuture.set(response);
+                } catch (Exception e) {
+                    rpcFuture.setException(e);
+                }
+            }
+
+            @Override
+            public void onErrorReceived(TResponse tResponse) {
+                rpcFuture.setException((Exception) tResponse.get(Names.EXCEPTION));
```
abhinav: This seems risky. Any way to avoid this?

yarpc-core/src/main/java/io/yarpc/exception/OutboundNotFoundException.java
```diff
@@ -0,0 +1,32 @@
...
+package io.yarpc.exception;
+
+public class OutboundNotFoundException extends Exception {
+    public OutboundNotFoundException(String message) {
+        super(message);
+    }
+
+    public OutboundNotFoundException(Exception e) {
+        super(e);
+    }
```
abhinav: If you remove all these constructors, you'll get the default exception constructors:

- No arguments
- Just the message
- Just the cause
- Both, the message and the cause

yarpc-core/src/main/java/io/yarpc/exception/RPCException.java
```diff
@@ -25,4 +25,8 @@
     public RPCException(String message) {
         super(message);
     }
+
+    public RPCException(Exception e) {
+        super(e);
```
abhinav: Same here. Remove the explicit constructors.

yarpc-core/src/main/java/io/yarpc/encoding/Serializer.java
junchaowu: Yes. It will be used in Thrift. I just submit json first. I will submit rest.

yarpc-core/src/main/java/io/yarpc/exception/RPCException.java
abhinav: Also, does your IDE not complain about missing `serialVersionUID` on your exception classes? You can have it generate it for you.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
junchaowu: The reason to use anonymous class here is to share the context. The responseFuture is per call base. Using the anonymous listener here can allow it not saving the future, but remember it.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
junchaowu: Class.newInstance() is common style here. This is just the beginning, as you think of future we add more configuration for the client , like client based configuration. It is quite useful to put a complex constructor into a helper method.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
junchaowu: what's risky?

yarpc-core/src/main/java/io/yarpc/exception/RPCException.java
junchaowu: no complain.

yarpc-core/src/main/java/io/yarpc/exception/OutboundNotFoundException.java
junchaowu: I don't understand what you try to say here. When I remove all these constructors. I can't build.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
abhinav: Static constructors/factory methods are a common pattern, yes. I have two issues here:

* `newInstance` is not a very good name for this method. Usually these methods have descriptive names based on what they're doing. Like, `fromConfig`, `newFromConfig`, etc.
* These methods make sense when there's some involved behavior that you want there, especially when that behavior doesn't directly tie into the arguments for that class' constructor. Like, for `fromConfig`, you would take a `Config` object, pull out the different attributes, validate them if necessary, and call the private constructor with the actual attributes you pulled out of the config/calculated based on what was in the Config. In this case, there's no behavior attached to it and it's just passing the argument to the regular constructor. That suggests there's no need for this static constructor. A case where you would need one is for example, `JsonClient.forService(RPC rpc, String serviceName)`, if the JsonClient was scoped to a single service.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
abhinav: Get and cast from map

yarpc-core/src/main/java/io/yarpc/exception/OutboundNotFoundException.java
abhinav: When you don't specify any explicit constructors for a class, it inherits those of the parent class. So, in this case, it would inherit these: https://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
abhinav: Yes, I understand why you need a nested class. I'm saying that you can move it to a separate internal class. Instead of,

    final String foo = ..
    new SomeInterface() {
        void something() {
            doStuff(foo);
        }
    }

You can do,

    private static class Foo implements SomeInterface {
        private final String foo;

        public Foo(String foo) {
            this.foo = foo;
        }

        void something() {
            doStuff(this.foo);
        }
    }

    // and where you had the anonymous class:

    new Foo(this.foo)

This makes the dependency on `this.foo` explicit rather than implicit. You can also use a non-static nested class if you want free access to the attributes of the class it's contained in but that's usually not necessary if you need to propagate only 2-3 attributes.

Also, note that your IDE most likely will do this for you automatically if you go into refactoring options.

yarpc-core/src/main/java/io/yarpc/exception/OutboundNotFoundException.java
abhinav: Okay, turns out that the constructors won't be inherited after all. Generate all 4 default constructors for all exceptions that don't have custom attributes.

yarpc-core/src/main/java/io/yarpc/exception/RPCException.java
junchaowu: discussed offline. Keep the constructors.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
anson627: newInstance is pretty common in java:
https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#newInstance--

yarpc-core/src/main/java/io/yarpc/Response.java
anson627: T, K, V that's java generic common practice.

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
anson627: let's not waste time on coding style, java is not python, space/tab won't screw up anything.

anson627:
:+1: let's get sth working quickly, instead of wasting time on naming convention or coding style. We are not google.

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
abhinav: Coding style consistency is not a waste of time. It contributes to the readability and ultimately, the maintainability of the code. If you want more than one person to work on the code, you need to have a consistent style throughout the codebase.

This comment is not about tabs or spaces. @junchaowu can use whichever he prefers. It was a nit about the line width. (Being a nit makes it low priority anyway.)

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
abhinav: That's an instance method on the `Class` type, not a static constructor of `Class` instances. Completely different.

I'm asking for a better name for the static constructor (which is indeed a common pattern), and whether you actually need a static constructor in this case at all.

The default should not be "use a static constructor". The default should be "use a normal constructor", and then if there is need for a static constructor, use one. Examples of where you might want a static constructor:

* You need construct a complex object with lots of attributes from another object with lots of attributes, and you won't be storing that object on the instance. `RPC(Config)` would probably be a static constructor.
* You need to always call a method on the constructed instance after instantiation. Since it's generally a bad idea to call instance methods from the constructor, you do it from the static constructor *after* the object has been initialized properly.

breerly:
@anson627 thanks for input - we like that energy :+1: 

A little bit of investment upfront on naming and code styling can go a long ways - so I have to agree with @abhinav on this one.

@blampe @kriskowal what say you two?

yarpc-core/src/main/java/io/yarpc/RPCFuture.java
breerly: I'm ok with exposing `sync` as a convenient, and potentially familiar, way to call `void get`.

yarpc-core/src/main/java/io/yarpc/Response.java
blampe: +1 T is fine.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
blampe: I agree with @abhinav, no need for this method if we can just call `new JsonClient(rpc)`.

junchaowu: no as I said in above comment, the future Jsonclient construction is not
only taking rpc instance , it will have client level configuration as well
for a complex construction, it is better to have helper method to construct
class. This is not Python having kwargs arguments to init a instance.

breerly: I think for now just stick to the bare minimum and instantiate the darn thing - we have a good amount of time before going beta, lets see if `newInstance` becomes necessary by then.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -0,0 +1,86 @@
...
+import io.yarpc.transport.ResponseListener;
+import io.yarpc.transport.TRequest;
+import io.yarpc.transport.TResponse;
+import io.yarpc.transport.Transformer;
+
+public class JsonClient extends RPCClient {
+    private static Serializer serializer = new JsonSerializer();
+    private final RPC rpc;
+
+    public JsonClient(RPC rpc) {
```
abhinav: IMO the client should depend on a single Outbound only rather than resolving the Outbound per-request from the RPC. Worth discussing elsewhere whether the low level request needs the remote service's name.

yarpc-core/src/main/java/io/yarpc/transport/TRequest.java
abhinav: Why does this have T in the name? You can call this `Request`. You have namespaces. In a context where you need to deal with this Request object and the other Request object, you import only one of them and use a fully qualified name for the other. 

Also, +1 on Grayson's comment. You have a generic `Request<T>`. The transport level request should just be `Request<byte[]>` or `Request<InputStream>` or `Request<ByteBuffer>`.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPClientInitializer.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public HTTPClientInitializer(ResponseListener responseListener) {
+        this.responseLister = responseListener;
+    }
+
+    @Override
+    protected void initChannel(Channel channel) throws Exception {
+        channel.pipeline().addLast("codec", new HttpClientCodec());
+        channel.pipeline().addLast("aggegator", new HttpObjectAggregator(Integer.MAX_VALUE));
+        channel.pipeline().addLast("customer", new HTTPClientInboundHandler(this.responseLister));
```
abhinav: Nit: You can chain calls to pipeline methods:

    channel.pipeline()
        .addLast(...)
        .addLast(...)
        .addLast(...);

yarpc-core/src/main/java/io/yarpc/transport/TResponse.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package io.yarpc.transport;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Response representation in transport layer.
+ */
+public class TResponse {
```
abhinav: Same issue as TRequest.

yarpc-core/src/main/java/io/yarpc/handler/BaseInboundHandler.java
```diff
@@ -19,38 +19,44 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
 import java.util.List;
 
-public class JSONInboundHandlerWrapper implements HandlerWrapper {
-    private final Logger logger = LoggerFactory.getLogger(JSONInboundHandlerWrapper.class);
-    private final ObjectMapper mapper;
+public abstract class BaseInboundHandler implements HandlerWrapper {
```
abhinav: Can you rename the wrapper interface too?

Also, once you get rid of the TRequest and TResponse types, you can also drop the HandlerWrapper interface since it's just `Handler<[]byte, []byte>` (or another equivalent to represent raw bytes). The encoding type can be transferred around in a different way.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
junchaowu: Let's discuss offline about this.

yarpc-core/src/main/java/io/yarpc/handler/BaseInboundHandler.java
junchaowu: When we have better idea of context, meta, headers and related things. We can come up better ideas.

Issue #18, opened 2016-02-17
Delete TRequest, TResponse, and HandlerWrapper


abhinav:
Follow up to https://github.com/yarpc/yarpc-java/pull/14

breerly:
:+1: here

junchaowu:
When we have better understand with out relationships among context propagation, meta, headers, body and so on, we will modify the TReq/TResp.

junchaowu:
done

Issue #11, opened 2016-02-11
Rename send to call


breerly:
The other langs use the term `call` when sending an `RPC` - lets do that here too.

junchaowu:
https://github.com/yarpc/yarpc-java/pull/29

PR #19, opened 2016-02-17
Add raw http support

junchaowu:
Add the raw http outbound and inbound. It works exact the same as Json.


yarpc-core/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -0,0 +1,87 @@
...
+public class RawClient extends RPCClient {
+    private static Serializer serializer = new RawSerializer();
+    private final RPC rpc;
+
+    public RawClient(RPC rpc) {
+        this.rpc = rpc;
+    }
+
+    @Override
+    public <I, O> RPCFuture<Response<O>> send(Request<I> request, final Class<O>
```
breerly: Please rename this to `call` to match the other languages.

yarpc-example/src/main/java/io/yarpc/example/http/raw/Client.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public static void main(String[] args) throws InterruptedException {
+        RPC rpc = new RPC(new Config(
+                "testServer",
+                new HashMap<String, Outbound>() {{
+                    try {
+                        put("rawServer", new HTTPOutbound("127.0.0.1", 9000));
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
```
breerly: can you explain why this try catch is necessary?

yarpc-example/src/main/java/io/yarpc/example/http/raw/Client.java
```diff
@@ -0,0 +1,65 @@
...
+                }}));
+
+        RPCClient client = new RawClient(rpc);
+        Request request = new Request.Builder<String>()
```
breerly: I'm really liking this request builder :+1:

breerly:
:shipit: 

(please circle back and address #11)

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -0,0 +1,87 @@
...
+            responseBodyType)
+            throws Exception {
+        Outbound outbound = this.rpc.lookUpOutbound(request.getSerivce());
+        request.setEncodingType(EncodingType.RAW);
+        TRequest tRequest = Transformer.toTRequest(request, RawClient.serializer);
+        final RPCFuture<Response<O>> rpcFuture = new RPCFuture<Response<O>>();
```
abhinav: I think you forgot to make this non-final when you moved the anonymous class to a named class.

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawClient.java
junchaowu: https://github.com/yarpc/yarpc-java/issues/11 will cover this.

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -0,0 +1,87 @@
...
+            this.responseBodyType = responseBodyType;
+            this.rpcFuture = rpcFuture;
+        }
+
+        @Override
+        public void onResponseReceived(TResponse tResponse) {
+
+            try {
+                O respBody = RawClient.serializer.unmarshal(
+                        (byte[]) tResponse.get(Names.BODY), responseBodyType);
```
prashantv: Why do we have an arbitrary bag of properties for `TResponse`? Why can't it be a class that has fields for the things we care about (headers, body, errors)

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawInboundHandler.java
```diff
@@ -0,0 +1,43 @@
...
+    public RawInboundHandler(Handler handler) {
+        super(new RawSerializer(), handler);
+    }
+
+    @Override
+    public String getEncodingType() {
+        return EncodingType.RAW;
+    }
+
+
```
prashantv: nit: why so many blank lines?

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawSerializer.java
```diff
@@ -0,0 +1,35 @@
...
+ * THE SOFTWARE.
+ */
+package io.yarpc.encoding.raw;
+
+import io.netty.util.CharsetUtil;
+import io.yarpc.encoding.Serializer;
+
+public class RawSerializer implements Serializer {
+    public <T> byte[] marshal(T value) {
+        return value.toString().getBytes(CharsetUtil.UTF_8);
```
prashantv: If `T` is `byte[]` will this work? How do you expect raw to be used?

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawClient.java
kriskowal: I wonder whether `RPCFuture<Response<O>>` can be simplified to `Response<O>`.

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawSerializer.java
kriskowal: Indeed, raw shouldn’t even necessarily be generic. Bytes for bytes.

yarpc-example/src/main/java/io/yarpc/example/http/raw/Client.java
```diff
@@ -0,0 +1,65 @@
...
+                .body("fruit")
```
kriskowal: I think the body should be bytes for raw. Using the raw "encoding" forfeits the responsibility of encoding and decoding strings.

yarpc-example/src/main/java/io/yarpc/example/http/raw/Server.java
```diff
@@ -0,0 +1,62 @@
...
+        RPC rpc = new RPC(new Config(
+                "testServer",
+                new ArrayList<Inbound>() {{
+                    add(new HTTPInbound(9000));
+                }}));
+
+        rpc.register("hello", new RawInboundHandler(new Handler<String, String>() {
+            @Override
+            public Response<String> handle(Request<String> request) throws RPCException {
+                return new Response.Builder<String>().body("apple").headers(request.getHeaders()).build();
```
kriskowal: Please break across lines. I’m with @malandrew on the alergy to fluent API’s. I would somewhat rather boring code here that constructs and populates a response object directly.

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawClient.java
junchaowu: This is due to at that moment we don't have a clear idea on what's the context, meta, and headers  and body 4 things look like. As I said in Json diff, it is temporary solution at that time. But the task has been created. https://github.com/yarpc/yarpc-java/issues/18  Will replace it with better structure.

yarpc-example/src/main/java/io/yarpc/example/http/raw/Client.java
junchaowu: That's another option to make rate just take bytes. I will open another diff to change it from string to byte.

yarpc-example/src/main/java/io/yarpc/example/http/raw/Server.java
breerly: I think this builder pattern is common when there are fields that are optional, possibly being filled in later by another layer.

abhinav: I'm like to avoid fluent setters on normal objects but fluent builders are a good idea for complex objects. They guarantee that you're not initializing an object with inconsistent internal state while keeping things readable. Fluent builders are the closest you can get to to named arguments in Java.

PR #20, opened 2016-02-17
config logger to use logback

junchaowu:


breerly:
Can you provide a description w the motivation for this change?

junchaowu:
Before this patch : when I run the example, it will show 
```
log4j:WARN No appenders could be found for logger (io.netty.util.internal.logging.InternalLoggerFactory).
log4j:WARN Please initialize the log4j system properly.
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```
I won't be able to see the log content because of this. After the patch, I can see the log in console.

breerly:
Thx :shipit:

PR #22, opened 2016-02-18
Thrift encoding

junchaowu:


yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
```diff
@@ -0,0 +1,87 @@
...
+import io.yarpc.encoding.EncodingType;
+import io.yarpc.encoding.Serializer;
+import io.yarpc.transport.Outbound;
+import io.yarpc.transport.ResponseListener;
+import io.yarpc.transport.TRequest;
+import io.yarpc.transport.TResponse;
+import io.yarpc.transport.Transformer;
+
+public class ThriftClient extends RPCClient {
+    private static Serializer serializer = new ThriftSerializer();
```
breerly: does the serializer have state? are you passing this instance around or just calling it from here?

Suggest making the class static and just calling it when you need.

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public ThriftClient(RPC rpc) {
+        this.rpc = rpc;
+    }
+
+    @Override
+    public <I, O> RPCFuture<Response<O>> send(Request<I> request, final Class<O>
+            responseBodyType)
+            throws Exception {
+        Outbound outbound = this.rpc.lookUpOutbound(request.getSerivce());
```
breerly: typo, you want `request.getService()`

breerly: a test should have caught this guy

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
```diff
@@ -0,0 +1,87 @@
...
+            this.rpcFuture = rpcFuture;
+        }
+
+        @Override
+        public void onResponseReceived(TResponse tResponse) {
+
+            try {
+                O respBody = ThriftClient.serializer.unmarshal(
+                        (byte[]) tResponse.get(Names.BODY), responseBodyType);
+                Response response = new Response.Builder<O>()
```
breerly: is your response immutable? Why not just overwrite the body?

breerly: oh yeah, because you're going from tresponse to response... that'll be updated soon right

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
```diff
@@ -0,0 +1,44 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.encoding.thrift;
+
+import io.yarpc.encoding.Serializer;
+import org.apache.thrift.TBase;
+import org.apache.thrift.TDeserializer;
+import org.apache.thrift.TSerializer;
+import org.apache.thrift.protocol.TBinaryProtocol;
```
breerly: I've been wondering if we should wrap apache's cli and library in our own `thriftrw` library... that would buy us the option of writing our own in the future without having to migrate anything... @abhinav is that crazy talk?

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class ThriftSerializer implements Serializer {
+    public <T> byte[] marshal(T value) throws Exception {
+        TSerializer serializer = new TSerializer(new TBinaryProtocol.Factory());
+        return serializer.serialize((TBase) value);
+    }
+
+    public <T> T unmarshal(byte[] value, Class<? extends T> clazz) throws Exception {
+        T base = clazz.newInstance();
```
breerly: `s/clazz/class`

yarpc-example/src/main/java/io/yarpc/example/http/thrift/Client.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.HashMap;
+
+public class Client {
+
+    public static void main(String[] args) throws InterruptedException {
+        RPC rpc = new RPC(new Config(
+                "testServer",
+                new HashMap<String, Outbound>() {{
+                    try {
+                        put("thriftServer", new HTTPOutbound("127.0.0.1", 9000));
```
breerly: Why do these outbounds throw `IOException`'s on instantiation? I think we should expect no side-effects when an object is instantiated - that would at least create a nicer interaction here (which we'll need when we get to file-based configuration).

yarpc-example/src/main/java/io/yarpc/example/http/thrift/Client.java
```diff
@@ -0,0 +1,80 @@
...
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
+                }}));
+
+        RPCClient client = new ThriftClient(rpc);
+        Request request = new Request.Builder<KeyValue.setValue_args>()
+                .body(new KeyValue.setValue_args("fruit", "banana"))
```
breerly: this `new KeyValue.SetValue_args(...)` makes me really really want to wrap Apache Thrift before 1.0 cc @abhinav

yarpc-example/src/main/java/io/yarpc/example/http/thrift/Client.java
```diff
@@ -0,0 +1,80 @@
...
+                .service("thriftServer")
+                .procedure("KeyValue::setValue")
```
breerly: Can we infer this from the body? Ideally this is optional for thrift requests

yarpc-example/src/main/java/io/yarpc/example/http/thrift/GetKeyValueHandler.java
```diff
@@ -0,0 +1,53 @@
...
+import io.yarpc.Request;
+import io.yarpc.Response;
+import io.yarpc.exception.RPCException;
+import io.yarpc.handler.Handler;
+import io.yarpc.thrift.generated.KeyValue;
+import io.yarpc.thrift.generated.NotFoundError;
+
+import java.util.Map;
+
+public class GetKeyValueHandler implements Handler<KeyValue.getValue_args, KeyValue.getValue_result> {
```
breerly: Nice, this handler is looking crisp.

yarpc-example/src/main/java/io/yarpc/example/http/thrift/SetKeyValueHandler.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private final Map<String, String> keyValueStore;
+
+    public SetKeyValueHandler(Map<String, String> keyValueStore) {
+        this.keyValueStore = keyValueStore;
+    }
+
+    @Override
+    public Response<KeyValue.setValue_result> handle(Request<KeyValue.setValue_args> request) throws RPCException {
+        String key = request.getBody().getKey();
```
breerly: Is it possible to build this `Request` by hand and pass it in to this handler manually for unit tests?

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
prashantv: `class` is a reserved word, so usually `clazz` is used

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
```diff
@@ -0,0 +1,44 @@
...
+
+        TDeserializer deserializer = new TDeserializer(new TBinaryProtocol.Factory());
+        deserializer.deserialize((TBase) base, value);
```
prashantv: can't you do `<T extends TBase>` so that you don't need casts here and above?

yarpc-example/src/main/java/io/yarpc/example/http/thrift/GetKeyValueHandler.java
```diff
@@ -0,0 +1,53 @@
...
+    public Response<KeyValue.getValue_result> handle(Request<KeyValue.getValue_args> request) throws RPCException {
+        String key = request.getBody().getKey();
+        String value = this.keyValueStore.get(key);
+
+        NotFoundError err = null;
+        if (value == null) {
+            err = new NotFoundError(key);
+        }
+
+        return new Response.Builder<KeyValue.getValue_result>().body(new KeyValue.getValue_result(value, err)).build();
```
prashantv: can we add static methods to help with construction of this so we don't need to specify the type `KeyValue.getValue_result` twice?

yarpc-example/src/main/java/io/yarpc/example/http/thrift/Server.java
```diff
@@ -0,0 +1,57 @@
...
+    public static void main(String[] args) throws InterruptedException {
+        RPC rpc = new RPC(new Config(
+                "testServer",
+                new ArrayList<Inbound>() {{
+                    add(new HTTPInbound(9000));
+                }}));
+
+        Map<String, String> keyValueStore = new ConcurrentHashMap<>();
+        rpc.register("KeyValue::setValue", new ThriftInboundHandler(new SetKeyValueHandler(keyValueStore)));
+        rpc.register("KeyValue::getValue", new ThriftInboundHandler(new GetKeyValueHandler(keyValueStore)));
```
prashantv: (for future) we should make it easy for the user to register a handlers for all these (ideally without having to specify strings for each method)

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
junchaowu: This serializer is shared by client and server. Do you remember in Json http, we have this serializer too.

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
junchaowu: turns out all the places are using serivce. Good catch.

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
junchaowu: Yeah I am going to do the delete the TResponse task for all.

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
junchaowu: +1 class is the keyword in Java

yarpc-example/src/main/java/io/yarpc/example/http/thrift/Client.java
junchaowu: My understanding about the situation is right now we only send out the thrift body not the whole envelop. The body ``setValue_args`` doesn't contain the service, and method name information.

yarpc-example/src/main/java/io/yarpc/example/http/thrift/SetKeyValueHandler.java
junchaowu: It is possible. Thrift request is no different with Json request. You can create it by ``Request<KeyValue.setValue_args>.Builder().body(new setValue_args(...))``

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
junchaowu: I can do it if this interface is just for Thrift. Since ``Serializer`` interface provides the same ``marshal`` and ``unmarshal`` functions for raw, thrift, json. If I change ``<T extends TBase>``, it will break other two.

yarpc-example/src/main/java/io/yarpc/example/http/thrift/GetKeyValueHandler.java
junchaowu: What do you mean by this?

yarpc-example/src/main/java/io/yarpc/example/http/thrift/Server.java
junchaowu: Agree. Right now it is limited by the apache thrift. We should be able to have it automatically figure out the name by having our own thrift gen

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
```diff
@@ -0,0 +1,87 @@
...
+    private static Serializer serializer = new ThriftSerializer();
+    private final RPC rpc;
+
+    public ThriftClient(RPC rpc) {
+        this.rpc = rpc;
+    }
+
+    @Override
+    public <I, O> RPCFuture<Response<O>> send(Request<I> request, final Class<O>
+            responseBodyType)
```
abhinav: This looks like you have your IDE set to 80 columns? That's too little for Java.

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
```diff
@@ -0,0 +1,44 @@
...
+package io.yarpc.encoding.thrift;
+
+import io.yarpc.encoding.Serializer;
+import org.apache.thrift.TBase;
+import org.apache.thrift.TDeserializer;
+import org.apache.thrift.TSerializer;
+import org.apache.thrift.protocol.TBinaryProtocol;
+
+public class ThriftSerializer implements Serializer {
+    public <T> byte[] marshal(T value) throws Exception {
```
abhinav: Use,

    public <T extends TBase> byte[] marshal(T value) throws Exception

And you won't have to cast to TBase.

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
```diff
@@ -0,0 +1,44 @@
...
+        TSerializer serializer = new TSerializer(new TBinaryProtocol.Factory());
+        return serializer.serialize((TBase) value);
+    }
+
+    public <T> T unmarshal(byte[] value, Class<? extends T> clazz) throws Exception {
```
abhinav: Same here

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
abhinav: To solve the serializer problem, you might be able to do:

    interface Serializer<A> {
        public <T extends A> byte[] marshal(T value) throws Exception;
        public <T extends A> T unmarshal(byte[], Class<? extends T> clazz) throws Exception;
    }

And change Json serializer to

    class JsonSerializer implements Serializer<Object> { ... }

And Thrift to

    class JsonSerializer implements Serializer<TBase> { ... }

yarpc-example/src/main/java/io/yarpc/example/http/thrift/Client.java
```diff
@@ -0,0 +1,80 @@
...
+                    try {
+                        put("thriftServer", new HTTPOutbound("127.0.0.1", 9000));
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
+                }}));
+
+        RPCClient client = new ThriftClient(rpc);
+        Request request = Request.Builder
+                .forBody(new KeyValue.setValue_args("fruit", "banana"))
```
abhinav: I assume we'll be using some code generation to make this `KeyValue.setValue(..)` or `keyValueClient.setValue(..)`?

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
prashantv: +1

abhinav:
Looks good in general. Couple minor comments, especially 
that cast-to-TBase thing.

prashantv:
lgtm, +1 for the cast thing.

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftSerializer.java
junchaowu: sure

yarpc-example/src/main/java/io/yarpc/example/http/thrift/Client.java
junchaowu: Yeah.

junchaowu:
I am going to merge this. I will make changes (add constrains to generic type, use bytebuffer to raw)  based on it for all three encoding types.

PR #29, opened 2016-02-23
Change Raw encoding to use Bytebuffer

junchaowu:
-Change the encodingType to encoding
-Add generic constrains into the RPC client and serializer
-Change send method name to call
-Add this to all instance method or variable references.
-Fix minor issues


yarpc-core/src/main/java/io/yarpc/RPCFuture.java
```diff
@@ -48,6 +48,6 @@ public boolean setException(Throwable throwable) {
     }
 
     public void sync() throws ExecutionException, InterruptedException {
-        get();
+        this.get();
```
anson627: not sure what this function do? a wrapper on top of one line code?

anson627:
lgtm, one general recommendation, try to avoid generic, unless have to. It makes both library users and authors' lives.

breerly:
Tests are failing - also, looks like you'll want these to be separate PR's for each item here

junchaowu:
abandon this. Will submit into different small PR's

PR #30, opened 2016-02-24
Change encodingType to encoding

junchaowu:
@yarpc/yarpc


breerly:
Nice. :sheep: once build passes.

yarpc-core/src/main/java/io/yarpc/Request.java
```diff
@@ -66,12 +66,16 @@ public void setHeaders(Map<String, String> headers) {
         this.headers = headers;
     }
 
-    public String getEncodingType() {
-        return this.encodingType;
+    public Class<T> getGenericType() {
```
prashantv: why this change?

prashantv:
assume `getGenericType` can be removed from this change. lgtm otherwise.

I'd also like us to consider moving away from a string to an enum for encoding.

junchaowu:
We had this discussion before. The reason we use the encoding as String is for extensibility purpose. In case in the future if user wants to add a custom encoding type. They can implement it without changing the encoding enum class. @abhinav

breerly:
@junchaowu possible to provide an enum to more easily interact with built-in encodings? we could track in another diff if you think that's a good idea.

prashantv:
@junchaowu Does the library allow other encodings to exist outside of the library and be plugged in?

junchaowu:
Our previous consideration is let's say in the future if user wants a new type of encoding, they can easy write their own without changing the encoding enum class. If users write their own serializer and client and server handler, they can use a custom encoding tyep.

yarpc-core/src/main/java/io/yarpc/encoding/Encoding.java
```diff
@@ -22,7 +22,7 @@
 package io.yarpc.encoding;
 
 
-public class EncodingType {
+public class Encoding {
```
abhinav: I would like to see this class go away and make something like `String getName()` part of the encoding interface. Basically you don't want any central place where you are enumerating the encodings.

breerly: +1

PR #32, opened 2016-02-24
Change raw type from String to ByteBuffer

junchaowu:
@yarpc/yarpc


yarpc-core/src/main/java/io/yarpc/RPCClient.java
```diff
@@ -22,7 +22,11 @@
 package io.yarpc;
 
 
-public abstract class RPCClient {
-    public abstract <I, O> RPCFuture<Response<O>> send(
+public abstract class RPCClient<B> {
+    public abstract <I extends B, O extends B> RPCFuture<Response<O>> call(
             Request<I> request, Class<O> responseBodyType) throws Exception;
+
+    public <I extends B> RPCFuture<Response<I>> call(Request<I> request) throws Exception {
```
junchaowu: This interface is created for the any encoding type which the response body type is same of the request body type. For instance, raw. The request and response is both ByteBuffer. So the ``client.call(request)`` is more convenient for users compared to ``client.call(request, ByteBuffer.class)``

breerly: I can buy that.

yarpc-core/src/main/java/io/yarpc/RPCClient.java
```diff
@@ -22,7 +22,11 @@
...
+        throw new UnsupportedOperationException("Please specific the response type.");
```
breerly: Will want to document this in the yarpc.io repo - can you create a todo to list out common library-level errors?

yarpc-example/src/main/java/io/yarpc/example/http/raw/Client.java
```diff
@@ -49,18 +50,18 @@ public static void main(String[] args) throws InterruptedException {
                     }
                 }}));
 
-        RPCClient client = new RawClient(rpc);
-        Request request = new Request.Builder<String>()
-                .body("fruit")
+        RawClient client = new RawClient(rpc);
+        Request<ByteBuffer> request = new Request.Builder<ByteBuffer>()
+                .body(ByteBuffer.wrap("world".getBytes()))
```
breerly: Nice - this interaction looks pretty sweet.

breerly:
I'm on-board :+1:

breerly:
:shipit:

yarpc-core/src/main/java/io/yarpc/RPCClient.java
abhinav: I'm confused. Why does this overloaded method exist? And why isn't it just abstract instead of raising UnsupportedOperationException?

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -47,15 +47,15 @@ public JsonClient(RPC rpc) {
     }
 
     @Override
-    public <I, O> RPCFuture<Response<O>> send(Request<I> request, final Class<O>
-            responseBodyType) throws Exception {
+    public <I extends Object, O extends Object> RPCFuture<Response<O>> call(
+            Request<I> request, final Class<O> responseBodyType) throws Exception {
```
abhinav: nit: why `final` on the parameter? You don't seem to have any anonymous classes here.

yarpc-core/src/main/java/io/yarpc/RPCClient.java
junchaowu: I put an implementation on the base class because by default I want the subclass doesn't need to care about implementing this method or not. If subclass wants an special implementation about it, then they can override it.

abhinav:
:+1:

yarpc-core/src/main/java/io/yarpc/RPCClient.java
kriskowal: s/specific/specify/

breerly: +1 to spelling correction here

PR #37, opened 2016-02-25
Add tests to increase test coverage.

junchaowu:
- Add unit tests for serializers and encoding handlers
- Add integration tests for three encoding types.


junchaowu:
@yarpc/java @yarpc/yarpc

breerly:
What does this increase the code cov to? I'll review shortly

junchaowu:
Right now we are 4% and it increases to the number in the snapshot below.
image

anson627:
:100:

breerly:
:ship:

yarpc-core/src/test/java/io/yarpc/encoding/json/JsonSerializerTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @Before
+    public void setup() {
+        this.jsonSerializer = new JsonSerializer();
+    }
+
+    @Test
+    public void testMarshal() throws JsonProcessingException {
+        Assert.assertTrue(Arrays.equals(
+                this.jsonSerializer.marshal(new TestQuery("test string"))
+                , new byte[]{123, 34, 99, 34, 58, 34, 116, 101, 115, 116, 32, 115, 116, 114, 105, 110, 103, 34, 125}
```
abhinav: Is this "test string" as bytes? I would just do `"test string".getBytes()`.

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @BeforeClass
+    public static void setupServer() throws IOException {
+        final Inbound httpInbound = setupHttpServer();
+        final Outbound httpOutbound = setupHttpClient();
+        rpc = new RPC(new Config(
+                "testServer",
+                new ArrayList<Inbound>() {{
+                    this.add(httpInbound);
+                }},
```
abhinav: You can use [`Arrays.asList`](https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#asList(T...)) instead.

    Arrays.asList(httpInbound)

yarpc-core/src/test/java/io/yarpc/encoding/json/JsonSerializerTest.java
junchaowu: This is how the TestQuery object is serialized as Json. so it is not same as "test string".getBytes()

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
```diff
@@ -0,0 +1,112 @@
...
+                new HashMap<String, Outbound>() {{
+                    this.put("testHttpServer", httpOutbound);
+                }}));
```
abhinav: Optional: Since you're using Guava, you can replace this with [`ImmutableMap.of`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/ImmutableMap.html#of())

    ImmutableMap.of("testHttpServer", httpOutbound)

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+
+public class RawIntegrationTest {
+    private static final String respString = "apple";
```
abhinav: Minor nit: I would set this as part of `@Before setUp()` (not `@BeforeClass` or static) and possibly make it a randomly generated string.

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
```diff
@@ -0,0 +1,112 @@
...
+    private static void registerEndpoint() {
+        registerHttpEndpoint();
+    }
+
+    private static void registerHttpEndpoint() {
+        RawInboundHandler rawInboundHandler = new RawInboundHandler(new Handler<ByteBuffer, ByteBuffer>() {
+
+            @Override
+            public Response<ByteBuffer> handle(Request<ByteBuffer> request) {
+                return Response.Builder.forBody(request.getBody()).build();
```
abhinav: Looks like this is a echo handler. Consider putting it in a named `RawEchoHandler` class or just adding a comment that this is an echo handler.

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Test
+    public void httpTest() throws Exception {
+        RawClient client = new RawClient(rpc);
+        Request request = new Request.Builder<ByteBuffer>()
+                .body(ByteBuffer.wrap(respString.getBytes()))
+                .service("testHttpServer")
+                .procedure("testHttpEndpoint")
+                .caller("testClient")
+                .timeout(new Duration(1000))
```
abhinav: nit: `new Duration(1000)` is the same as `Duration.millis(1000)` but the latter makes it clear that we're dealing with milliseconds.

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawSerializerTest.java
```diff
@@ -19,21 +19,37 @@
...
+    @Before
+    public void setup() {
+        this.rawSerializer = new RawSerializer();
+    }
+
+    @Test
+    public void testMarshal() {
+        Assert.assertTrue(
+                Arrays.equals(this.rawSerializer.marshal(ByteBuffer.wrap("test string".getBytes()))
+                        , new byte[]{116, 101, 115, 116, 32, 115, 116, 114, 105, 110, 103})
```
abhinav: I would prefer `"test string".getBytes()` over hand-coded list of byte values.

yarpc-core/src/test/java/io/yarpc/encoding/thrift/ThriftInboundHandlerTest.java
```diff
@@ -0,0 +1,62 @@
...
+            public Response<KeyValue.getValue_result> handle(
+                    Request<KeyValue.getValue_args> request) throws RPCException {
+                return Response.Builder.forBody(new KeyValue.getValue_result("test", null)).build();
+            }
+        });
+    }
+
+    @Test
+    public void testParseThriftHandler() {
+
```
abhinav: nit: delete empty line

yarpc-core/src/test/java/io/yarpc/encoding/thrift/ThriftInboundHandlerTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Before
+    public void setup() {
+        this.thriftInboundHandler = new ThriftInboundHandler(new Handler<KeyValue.getValue_args,
+                KeyValue.getValue_result>() {
+
+            @Override
+            public Response<KeyValue.getValue_result> handle(
+                    Request<KeyValue.getValue_args> request) throws RPCException {
+                return Response.Builder.forBody(new KeyValue.getValue_result("test", null)).build();
+            }
```
abhinav: Out of scope for this diff: You'll be able to test more cases in these tests if instead of hand-coding these endpoints, you used a mocking library. Check out EasyMock or Mockito.

yarpc-core/src/test/java/io/yarpc/encoding/thrift/ThriftIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+    @BeforeClass
+    public static void setupServer() throws IOException {
+        final Inbound httpInbound = setupHttpServer();
+        final Outbound httpOutbound = setupHttpClient();
+        rpc = new RPC(new Config(
+                "testServer",
+                new ArrayList<Inbound>() {{
+                    this.add(httpInbound);
+                }},
+                new HashMap<String, Outbound>() {{
```
abhinav: Same thing as above about `Arrays.asList` and `ImmutableMap.of`.

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
junchaowu: The `@Before` will be called for each test case in the class. However as for integration test, I'd like to setup servers at beginning and close at end just once.

yarpc-core/src/test/java/io/yarpc/encoding/thrift/ThriftIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+import io.yarpc.RPCFuture;
+import io.yarpc.Request;
+import io.yarpc.Response;
+import io.yarpc.config.Config;
+import io.yarpc.exception.RPCException;
+import io.yarpc.handler.Handler;
+import io.yarpc.thrift.generated.KeyValue.getValue_args;
+import io.yarpc.thrift.generated.KeyValue.getValue_result;
+import io.yarpc.thrift.generated.KeyValue.setValue_args;
+import io.yarpc.thrift.generated.KeyValue.setValue_result;
```
abhinav: Nit: In the other class, you did not do a full import of `{get,set}Value_{args,result}`. You stopped at `KeyValue` and then referred to these as `KeyValue.getValue_args`/whatever. Pick one and be consistent please.

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawSerializerTest.java
junchaowu: Yes for raw they are the same. I will change it to getBytes

yarpc-core/src/test/java/io/yarpc/encoding/thrift/ThriftIntegrationTest.java
```diff
@@ -0,0 +1,136 @@
...
+        return new HTTPOutbound(ip, httpPort);
+    }
+
+    private static void registerEndpoint() {
+        registerHttpEndpoint();
+    }
+
+    private static void registerHttpEndpoint() {
+        ThriftInboundHandler getValueHandler =
+                new ThriftInboundHandler(new Handler<getValue_args, getValue_result>() {
```
abhinav: Nit: In the other class, you did not do a full import of `{get,set}Value_{args,result}`. You stopped at `KeyValue` and then referred to these as `KeyValue.getValue_args`/whatever. Pick one and be consistent please.

yarpc-core/src/test/java/io/yarpc/encoding/thrift/ThriftSerializerTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Before
+    public void setup() {
+        this.thriftSerializer = new ThriftSerializer();
+    }
+
+    @Test
+    public void testMarshal() throws Exception {
+        Assert.assertTrue(Arrays.equals(
+                this.thriftSerializer.marshal(new KeyValue.getValue_args("test string"))
+                , new byte[]{11, 0, 1, 0, 0, 0, 11, 116, 101, 115, 116, 32, 115, 116, 114, 105, 110, 103, 0}
```
abhinav: For hand-coded Thrift payloads, you want to document the sections. Here's what this is for this payload:

    11,  // type: binary
    0, 1,  // id: 1
    0, 0, 0, 11, // length: 11
    116, 101, 115, 116, 32, 115, 116, 114, 105, 110, 103,  // "test string",
    0, // end struct

Actually, this also brings up: For the other payloads, if you don't want to do `getBytes()`, you should at least document what the 50 random looking numbers represent so that someone can change them.

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
abhinav: I'm talking about the respString though. It's okay to generate that randomly before each test.

abhinav:
Yay tests. I mostly have nits about styling and stuff.

yarpc-core/src/test/java/io/yarpc/encoding/json/JsonSerializerTest.java
abhinav: Is it `"\"test string\"".getBytes()`?

yarpc-core/src/test/java/io/yarpc/encoding/thrift/ThriftSerializerTest.java
junchaowu: oh this is great. Thanks for pointing out.

yarpc-core/src/test/java/io/yarpc/encoding/thrift/ThriftIntegrationTest.java
junchaowu: done.

PR #42, opened 2016-03-01
User Bytebuffer request/response for transport layer 

junchaowu:
Use request<ByteBuffer> response<ByteBuffer>  for the general request/response representation in transport layer. @yarpc/java 

This one will replace the previous `Map<String, Object>` structure to store all the necessary information.


yarpc-core/src/main/java/io/yarpc/BaseRequest.java
```diff
@@ -0,0 +1,171 @@
...
+import io.yarpc.context.Context;
+import org.apache.commons.lang.builder.ToStringBuilder;
+import org.joda.time.Duration;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class BaseRequest {
+    protected Context context;
+    protected Map<String, String> headers;
```
breerly: nice

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -69,12 +67,12 @@ private JsonResponseListener(Class<O> responseBodyType, RPCFuture<Response<O>> r
         }
 
         @Override
-        public void onResponseReceived(TResponse tResponse) {
+        public void onResponseReceived(TransportResponse transportResponse) {
```
breerly: the naming on this is much better

yarpc-core/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -55,9 +53,9 @@ public RawClient(RPC rpc) {
         request.setEncoding(Encoding.RAW);
         logger.debug(String.format("sending request %s", request.toString()));
 
-        TRequest tRequest = Transformer.toTRequest(request, RawClient.serializer);
+        TransportRequest transportRequest = new TransportRequest(request, RawClient.serializer);
```
breerly: might make sense to have this class be injected... we can do that later though.

yarpc-core/src/main/java/io/yarpc/transport/TransportRequest.java
```diff
@@ -21,24 +21,49 @@
...
+    }
+
+    public TransportRequest(Request request, Serializer serializer) throws Exception {
+        super(request);
+        body = serializer.marshal(request.getBody());
+    }
 
-    public TRequest() {
-        this.attributes = new HashMap<>();
+    public byte[] getBody() {
```
breerly: :+1:

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPClientInboundHandler.java
```diff
@@ -46,7 +50,20 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
     protected void messageReceived(ChannelHandlerContext channelHandlerContext, FullHttpResponse httpResponse) throws
             Exception {
         // Convert to rpc context
-        this.responseListener.onResponseReceived(Transformer.toTResponse(httpResponse));
+        byte[] responseContent = new byte[httpResponse.content().readableBytes()];
+        httpResponse.content().readBytes(responseContent);
+        Map<String, String> headers = new HashMap<>();
```
breerly: Let's tasks for a "headers" Crossdock test.

breerly:
lgtm

yarpc-core/src/main/java/io/yarpc/BaseRequest.java
```diff
@@ -0,0 +1,171 @@
+/**
+ * Copyright (c) 2016, Uber Technologies, Inc
```
ShanniLi: Why do you need BaseRequest and BaseResponse? Why not to just put it to Request/Response?

junchaowu: this is because request/response is generic class, the transport request and response is using byte array which doesn't fit into generic. And the transport req/response share most of variables in the request and response. Therefore move the common part into base class.

ShanniLi:
lgtm

yarpc-core/src/main/java/io/yarpc/BaseRequest.java
```diff
@@ -0,0 +1,171 @@
...
+        private Duration nestedTimeout;
+        private String nestedCaller;
+
+        public Builder() {
+        }
+
+        /**
+         * The solution for the unchecked cast warning.
+         */
+        public abstract T getThis();
```
abhinav: Do you still need this? The `Builder<T extends Builder<T>>` thing should give you a self-referential `T`. Worst case, just add a `@SuppressWarnings("unchecked")` on `class Builder` rather than adding a code-level hack for a warning we know won't be a problem.

yarpc-core/src/main/java/io/yarpc/transport/TransportRequest.java
```diff
@@ -21,24 +21,49 @@
...
+import io.yarpc.Request;
+import io.yarpc.encoding.Serializer;
 
-/**
- * Request representation in transport layer.
- */
-public class TRequest {
-    private Map<String, Object> attributes;
+public class TransportRequest extends BaseRequest {
+    private byte[] body;
```
abhinav: Can you explain why you chose to make this a byte array rather than a byte buffer like the raw request? Doesn't converting it to a byte array lose you the no-copy property of ByteBuffer?

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPClientInboundHandler.java
```diff
@@ -46,7 +50,20 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
+
+        for (Map.Entry<CharSequence, CharSequence> entry : httpResponse.headers().entries()) {
+            headers.put(entry.getKey().toString(), entry.getValue().toString());
```
abhinav: Are your header keys case insensitive? Do you need to do conversion here?

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
```diff
@@ -82,20 +82,20 @@ public void stop() {
...
+        ByteBuf bbuf = Unpooled.wrappedBuffer(transportRequest.getBody());
 
         //TODO convert context property into request headers
-        httpRequest.headers().add(Names.PROCEDURE, (String) tRequest.get(Names.PROCEDURE));
-        httpRequest.headers().add(Names.ENCODING, (String) tRequest.get(Names.ENCODING));
-        httpRequest.headers().add(Names.TTLMS, String.valueOf(tRequest.get(Names.TTLMS)));
-        httpRequest.headers().add(Names.CALLER, (String) tRequest.get(Names.CALLER));
-        httpRequest.headers().add(Names.SERVICE, (String) tRequest.get(Names.SERVICE));
+        httpRequest.headers().add(Names.PROCEDURE, transportRequest.getProcedure());
+        httpRequest.headers().add(Names.ENCODING, transportRequest.getEncoding());
```
abhinav: yay no more bucket

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
```diff
@@ -82,20 +82,20 @@ public void stop() {
...
 
-    private void sendRequest(Channel channel, TRequest tRequest) throws Exception {
+    private void sendRequest(Channel channel, TransportRequest transportRequest) throws Exception {
         FullHttpRequest httpRequest = new DefaultFullHttpRequest(
                 HttpVersion.HTTP_1_1,
                 HttpMethod.POST,
                 PATH);
 
-        ByteBuf bbuf = Unpooled.wrappedBuffer((byte[]) tRequest.get(Names.BODY));
+        ByteBuf bbuf = Unpooled.wrappedBuffer(transportRequest.getBody());
```
abhinav: FWIW it appears that netty's ByteBuf works well with nio.ByteBuffer. Might be able to avoid copying by using that.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -66,6 +67,26 @@ protected void messageReceived(ChannelHandlerContext channelHandlerContext, Full
         this.forwardRequest(channelHandlerContext, httpRequest, handlerWrapper);
     }
 
+    private TransportRequest parseRequest(FullHttpRequest httpRequest) {
+        byte[] requestContent = new byte[httpRequest.content().readableBytes()];
+        httpRequest.content().readBytes(requestContent);
```
abhinav: You'll be able to use [`content().nioBuffer()`](http://netty.io/4.0/api/io/netty/buffer/ByteBuf.html#nioBuffer()) and pass that down to TransportReq/Res if you switch to ByteBuffer there as well without making copies of the body.

abhinav:
Overall looks good. Minor nits + I don't understand the decision to not use ByteBuffer.

breerly:
@junchaowu can you explain the motivation behind not using ByteBuffer? I'm a bit curious as well.

yarpc-core/src/main/java/io/yarpc/BaseRequest.java
junchaowu: I know. The purpose of this is to avoid that warning nothing hurt here.

abhinav: Yeah, I agree. I'm saying there's a much simpler way to get rid of the warning: Add `@SuppressWarnings("unchecked")` on the Builder class.

junchaowu:
Switch to ByteBuffer so that we don't need baserequest baseresponse any more. Get rid of additional transportrequest and transportresponse as well.

yarpc-core/src/main/java/io/yarpc/Request.java
```diff
@@ -22,37 +22,63 @@
...
+        this.encoding = builder.nestedEncoding;
         if (this.timeout == null) {
             this.timeout = Duration.millis(1000);
         }
         this.caller = builder.nestedCaller;
     }
 
-    public Class<T> getGenericType() {
-        return (Class<T>) this.body.getClass();
+    public static Request<ByteBuffer> toByteBufferRequest(Request request, Serializer serializer) throws Exception {
```
breerly: suggest `toByteBufferedRequest`

yarpc-core/src/main/java/io/yarpc/Request.java
```diff
@@ -22,37 +22,63 @@
...
+                .context(request.getContext())
+                .encoding(request.getEncoding())
+                .timeout(request.getTimeout())
+                .procedure(request.getProcedure())
+                .headers(request.getHeaders())
+                .service(request.getService())
+                .build();
+    }
+
+    public static Request fromByteBufferRequest(Request<ByteBuffer> request, Serializer serializer, Class<?> bodyType)
```
breerly: suggest `fromByteBufferedRequest`

breerly:
Big big improvement :shipit:

yarpc-core/src/main/java/io/yarpc/Request.java
abhinav: Use the full type here please. Don't use generic types without `<...>`.

```java
public static <T> Request<ByteBuffer> toByteBufferRequest(Request<T> request, Serializer<T> serializer)
```

Same for `fromByteBufferRequest`. Also, for naming, you could just call them "serializeRequestBody" and "deserializeRequestBody" since that's exactly what's changing.

abhinav: Also not for this PR: that `throws Exception` is too wide for just serialization. Let's resolve that in a different PR. Created #44.

yarpc-core/src/main/java/io/yarpc/Request.java
```diff
@@ -22,37 +22,63 @@
...
 
+import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Map;
 
 public class Request<T> {
+    protected Context context;
+    protected Map<String, String> headers;
+    protected String encoding;
+    protected String service;
```
abhinav: Since this is no longer inherited anywhere, you should convert these all back to private.

yarpc-core/src/main/java/io/yarpc/Request.java
```diff
@@ -126,14 +144,27 @@ public String toString() {
         return ToStringBuilder.reflectionToString(this);
     }
 
+    public Class<T> getGenericType() {
```
abhinav: Nit: What about `getBodyClass`?

yarpc-core/src/main/java/io/yarpc/Response.java
```diff
@@ -30,11 +32,32 @@
...
     private Response(Builder<T> builder) {
         this.headers = builder.nestedHeaders;
-        this.body = builder.nestedRespBody;
         this.context = builder.nestedContext;
+        this.body = builder.nestedRespBody;
+        this.exception = builder.nestedException;
+    }
+
+    public static Response<ByteBuffer> toByteBufferResponse(Response response, Serializer serializer) throws Exception {
+        return Response.Builder
```
abhinav: Same comments as above: Use fully populated types for the generics, and optionally rename to serializeRequestBody and deserializeRequestBody.

yarpc-core/src/main/java/io/yarpc/encoding/Serializer.java
```diff
@@ -21,8 +21,11 @@
  */
 package io.yarpc.encoding;
 
+import java.nio.ByteBuffer;
+
 public interface Serializer<B> {
-    <T extends B> byte[] marshal(T value) throws Exception;
+    <T extends B> ByteBuffer marshal(T value) throws Exception;
+
+    <T extends B> T unmarshal(ByteBuffer value, Class<T> clazz) throws Exception;
```
abhinav: Just leaving this here: https://github.com/yarpc/yarpc-java/issues/44

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonSerializer.java
```diff
@@ -25,22 +25,24 @@
...
 
 
 public class JsonSerializer implements Serializer<Object> {
     private final static Logger logger = LoggerFactory.getLogger(JsonSerializer.class);
 
-    public <T extends Object> byte[] marshal(T value) throws JsonProcessingException {
+    public <T extends Object> ByteBuffer marshal(T value) throws JsonProcessingException {
         ObjectMapper mapper = new ObjectMapper();
-        return mapper.writeValueAsString(value).getBytes(CharsetUtil.UTF_8);
+        return ByteBuffer.wrap(mapper.writeValueAsString(value).getBytes(CharsetUtil.UTF_8));
```
abhinav: You can use [`writeValueAsBytes`](https://fasterxml.github.io/jackson-databind/javadoc/2.0.0/com/fasterxml/jackson/databind/ObjectMapper.html#writeValueAsBytes(java.lang.Object)) here.

yarpc-core/src/main/java/io/yarpc/handler/BaseInboundHandler.java
```diff
@@ -47,12 +46,12 @@ public BaseInboundHandler(Serializer serializer, Handler handler) {
...
     @Override
-    public TResponse handle(TRequest tRequest) throws RPCException {
-        this.validate(tRequest);
+    public Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest) throws RPCException {
+        this.validate(transportRequest);
         try {
-            Response response = this.getHandler().handle(Transformer.toRequest(tRequest, this.serializer,
-                    getReqestBodyType()));
-            return Transformer.toTResponse(response, this.serializer);
+            Request request = Request.fromByteBufferRequest(transportRequest, this.serializer, requestBodyType);
```
abhinav: This class is using a lot of unpopulated generics. Fixing this is out of scope for this PR so I created https://github.com/yarpc/yarpc-java/issues/45 with an explanation of how it should probably happen.

yarpc-core/src/main/java/io/yarpc/handler/HandlerWrapper.java
```diff
@@ -21,12 +21,14 @@
...
+import io.yarpc.Response;
 import io.yarpc.exception.RPCException;
-import io.yarpc.transport.TRequest;
-import io.yarpc.transport.TResponse;
+
+import java.nio.ByteBuffer;
 
 public interface HandlerWrapper {
-    TResponse handle(TRequest tRequest) throws RPCException;
+    Response<ByteBuffer> handle(Request<ByteBuffer> tRequest) throws RPCException;
```
abhinav: We should delete this class in a different PR: https://github.com/yarpc/yarpc-java/issues/46

yarpc-core/src/main/java/io/yarpc/transport/Outbound.java
```diff
@@ -22,8 +22,12 @@
 package io.yarpc.transport;
 
 
+import io.yarpc.Request;
+
+import java.nio.ByteBuffer;
+
 public abstract class Outbound {
-    public abstract void send(TRequest tRequest, ResponseListener responseListener) throws Exception;
+    public abstract void send(Request<ByteBuffer> transportRequest, ResponseListener responseListener) throws Exception;
```
abhinav: small nit: when unambiguous, it's okay to use `request` rather than `transportRequest`

yarpc-core/src/main/java/io/yarpc/transport/ResponseListener.java
```diff
@@ -21,8 +21,12 @@
...
+import io.yarpc.Response;
+
+import java.nio.ByteBuffer;
+
 public interface ResponseListener {
-    void onResponseReceived(TResponse tResponse);
+    void onResponseReceived(Response<ByteBuffer> tResponse);
 
-    void onErrorReceived(TResponse tResponse);
+    void onErrorReceived(Response<ByteBuffer> tResponse);
```
abhinav: nit: `response` or `transportResponse` rather than `tResponse`

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPClientInboundHandler.java
```diff
@@ -45,8 +50,19 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
     @Override
     protected void messageReceived(ChannelHandlerContext channelHandlerContext, FullHttpResponse httpResponse) throws
             Exception {
-        // Convert to rpc context
-        this.responseListener.onResponseReceived(Transformer.toTResponse(httpResponse));
+        // TODO Convert to rpc context
+        Map<String, String> headers = new HashMap<>();
+
+        for (Map.Entry<CharSequence, CharSequence> entry : httpResponse.headers().entries()) {
+            headers.put(entry.getKey().toString(), entry.getValue().toString());
```
abhinav: Application headers are case insensitive. Created https://github.com/yarpc/yarpc-java/issues/47 to fix this later.

yarpc-core/src/main/java/io/yarpc/Request.java
abhinav: Actually, for the generic change, let's do that in a separate PR: https://github.com/yarpc/yarpc-java/issues/45

abhinav:
This is much better! Thanks.

I have a couple minor nits, and some other things that are out of scope for this PR (created GH issues).

LGTM besides that. :shipit:

abhinav:
:+1:

PR #51, opened 2016-03-04
Remove the version number from the maven build.

junchaowu:
Fix the example to use external ip.
Fix the cross dock to use the jar without worrying about the version number


breerly:
Looks great :+1:

breerly:
Looks like you'll want to update coveralls... it keeps flapping even though you didn't edit any java files

PR #54, opened 2016-03-07
Add TChannel integration test.

junchaowu:
Refactor some of the integration test structure for reuse of code.


yarpc-core/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -38,41 +41,58 @@
...
     private static final String respString = "apple";
-    private static final int httpPort = 9090;
-    private static final String ip = "127.0.0.1";
+    private static final int httpPort = 9091;
+    private static final int tchannelPort = 9092;
+    private static final String httpServer = "testHttpServer";
+    private static final String tchannelServer = "testTChannelServer";
+    private static final String endpoint = "testEndpoint";
+    private static final String clientName = "testClient";
     private static RPC rpc;
```
abhinav: Nit: this should be final too if you're not modifying it per-test. If you're modifying it per-test, it shouldn't be static.

yarpc-core/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -38,41 +41,58 @@
...
+    private static final String rpcName = "testRPC";
 
     @BeforeClass
     public static void setupServer() throws IOException {
-        final Inbound httpInbound = setupHttpServer();
+        final TChannel tchannel = new TChannel.Builder("json")
```
abhinav: Why is this final? I don't see it used in an anonymous class anywhere.

Same for the other local variables in this method.

yarpc-core/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -38,41 +41,58 @@
...
         rpc = new RPC(new Config(
-                "testServer",
-                new ArrayList<Inbound>() {{
-                    this.add(httpInbound);
-                }},
+                rpcName,
+                Arrays.asList(httpInbound, tchannelInbound),
                 new HashMap<String, Outbound>() {{
-                    this.put("testHttpServer", httpOutbound);
+                    this.put(httpServer, httpOutbound);
```
abhinav: Just create a local variable and `put` on it there.

yarpc-core/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -38,41 +41,58 @@
...
+                    this.put(tchannelServer, tchannelOutbound);
                 }}));
         registerEndpoint();
         rpc.start();
     }
 
-    private static Outbound setupHttpClient() throws IOException {
-        return new HTTPOutbound(ip, httpPort);
+    private static Inbound setupTChannelServer(TChannel tchannel) {
+        return new TChannelInbound(tchannel);
```
abhinav: This isn't used anywhere but once in setupServer, and it's a single line method. I would just delete the method and write `new TChannelInbound(tchannel)` up there.

yarpc-example/src/main/java/io/yarpc/example/tchannel/raw/Client.java
```diff
@@ -44,7 +45,9 @@ public static void main(String[] args) throws InterruptedException, IOException
         RPC rpc = new RPC(new Config(
                 "testServer",
                 new HashMap<String, Outbound>() {{
-                    this.put("rawServer", new TChannelOutbound(tchannel, "0.0.0.0", 9000));
+                    this.put(
+                            "rawServer",
+                            new TChannelOutbound(tchannel, Inet4Address.getLocalHost().getHostAddress(), 9000));
```
abhinav: Again, just create a local variable `outbounds` and do a `put` on that. You don't *have* to keep it all in a single expression. Prefer readability.

yarpc-example/src/main/java/io/yarpc/example/tchannel/thrift/Client.java
```diff
@@ -43,7 +44,9 @@ public static void main(String[] args) throws InterruptedException, IOException
         RPC rpc = new RPC(new Config(
                 "testServer",
                 new HashMap<String, Outbound>() {{
-                    this.put("thriftServer", new TChannelOutbound(tchannel, "0.0.0.0", 9000));
+                    this.put(
+                            "thriftServer",
+                            new TChannelOutbound(tchannel, Inet4Address.getLocalHost().getHostAddress(), 9000));
```
abhinav: Same as above

abhinav:
:ok:  once it passes

PR #56, opened 2016-03-08
Modify harness testing to be able to trigger in unit test

junchaowu:


yarpc-core/src/test/java/io/yarpc/crossdock/HarnessServerInboundHandler.java
```diff
@@ -70,7 +67,7 @@ protected void messageReceived(ChannelHandlerContext channelHandlerContext, Full
         String respBody;
         HarnessResponse harnessResponse;
         ObjectMapper objectMapper = new ObjectMapper();
-        if (behavior.equals("echo")) {
+        if (behavior.equals(EchoTest.name)) {
```
breerly: Perhaps we should have a map of:

```
{
    "echo": echo,
}
``` 

Then just attempt to pull the behavior out of the map.

yarpc-core/src/test/java/io/yarpc/crossdock/echo/EchoTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.UUID;
+
+/**
+ * EchoTest contains tests for 'echo' behavior
+ */
+public class EchoTest {
+    public static final String name = "echo";
+
+    public String testEcho(String server, int remotePort) throws Exception {
```
breerly: so this gets called from the test suite and crossdock test?

breerly: answered my own question :+1:

breerly:
lgtm

yarpc-core/src/test/java/io/yarpc/crossdock/HarnessServerInboundHandler.java
junchaowu: unless each test has unified interface, otherwise it is doable in java. Therefore until we have further understanding of the crossdock test cases, we may adapt it in the future.

kriskowal: In Node.js, each behavior needs a params map, a results collection to append to, and must "return a future". Something similar should work for Java, perhaps in a follow-up.

yarpc-core/src/test/java/io/yarpc/crossdock/echo/EchoTest.java
```diff
@@ -0,0 +1,42 @@
...
+        EchoClient echoClient = new EchoClient(server, remotePort);
+        String hash = UUID.randomUUID().toString().substring(0, 7);
+        Response<Echo> response = echoClient.call(server, new Echo(hash)).get();
+        Assert.assertEquals(response.getBody().getToken(), hash);
```
kriskowal: How does this assertion thread through to whether the harness receive "passed" or "failed"? In Node.js, I had to stop using the unit test assertion library for crossdock behaviors and instead implemented an EqualsAssertion that implements a TestResult (effectively), with the status and output properties produced based on the given constructor arguments. That result gets either tacked onto the results for the Crossdock HTTP response, or for unit tests, I assert that all of the test results are passed.

kriskowal:
:ok: Looks like progress. Let’s make sure that test failures get threaded through when the tests are not running in jUnit.

yarpc-core/src/test/java/io/yarpc/crossdock/echo/EchoTest.java
abhinav: I think assertion will throw an exception

junchaowu:
Double check the threaded when assert fails. It will catch the assertError.

PR #59, opened 2016-03-09
Use try with resource to avoid memory leak from tchannel response

junchaowu:
TChannel's response need to call release after it is used. Therefore use try with resource strategy to release the resources automatically and avoid memory leak.


anson627:
:+1:

PR #60, opened 2016-03-10
Move crossdock as a separate project.

junchaowu:
Reason:
The existing solution for supporting crossdock harness behaviors results all the cross dock harness code is built into the yarpc-core jar. And it is mixed between junit tests class and the test harness server class. The code path for crossdock and junit test are not clear at all. This makes the yarpc-core jar contains so many things that the normal developers don't need. It also causes some dependencies jars included into the yarpc-core jar.

Solution:
Make crossdock as a separate module of yarpc-java. Its responsibility is to setup the harness server and call the harness tests in the yarpc-core jar. The still share the code harness testing code but much clear.

- some related work is to optimize the maven dependencies to avoid duplicate and useless jars built into final jar file.


breerly:
:shipit: assuming you can still run crossdock behaviors from your regular test suite

junchaowu:
merged this to add move crossdock testing based on this.

PR #65, opened 2016-03-16
Switch maven to gradle

junchaowu:


junchaowu:
Here is the comparison between maven and gradle: http://gradle.org/maven_vs_gradle/

breerly:
Much XML deleted. :ship: once travis is happy.

junchaowu:
Coveralls is not working recently. I will skip it.

Issue #46, opened 2016-03-02
Delete HandlerWrapper


abhinav:
HandlerWrapper is defined as:

```java
public interface HandlerWrapper {
    Response<ByteBuffer> handle(Request<ByteBuffer> tRequest) throws RPCException;		  
    String getEncoding();
}
```

And Handler is defined as:

```java
public interface Handler<I, O> {
    Response<O> handle(Request<I> request) throws RPCException;
}
```

We should delete HandlerWrapper and just use `Handler<ByteBuffer, ByteBuffer>` in its place. Alternatively, we can rename it to `TransportHandler`. 

Worth noting: `HandlerWrapper.getEncoding()` is not used anywhere and that information is available via `Request.getEncoding()` anyway.

PR #66, opened 2016-03-21
remove handlerWrapper

junchaowu:
this is to fix https://github.com/yarpc/yarpc-java/issues/46


yarpc-core/src/main/java/io/yarpc/HandlerRegistry.java
```diff
@@ -22,12 +22,12 @@
 package io.yarpc;
 
 
-import io.yarpc.handler.HandlerWrapper;
+import io.yarpc.handler.Handler;
 
 public interface HandlerRegistry {
-    void addHandler(String procedure, HandlerWrapper handler);
+    void addHandler(String procedure, Handler handler);
```
abhinav: Don't use generics types without type parameters. Parameterize the function when possible. For cases where it's not possible, use `<?>` rather than no-parameter, but prefer to always have an actual parameter.

yarpc-core/src/main/java/io/yarpc/MapHandlerRegistry.java
```diff
@@ -22,16 +22,16 @@
...
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
 public class MapHandlerRegistry implements HandlerRegistry {
 
-    private final ConcurrentMap<String, HandlerWrapper> handlers = new ConcurrentHashMap<>();
+    private final ConcurrentMap<String, Handler> handlers = new ConcurrentHashMap<>();
 
-    public void addHandler(String procedure, HandlerWrapper handler) {
+    public void addHandler(String procedure, Handler handler) {
```
abhinav: Same as above

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
```diff
@@ -34,7 +34,6 @@ public JsonInboundHandler(Handler handler) {
         super(new JsonSerializer(), handler);
     }
 
-    @Override
     public String getEncoding() {
```
abhinav: If this isn't used anywhere, remove it

abhinav:
Please provide full type parameters for generics. Don't use generics without type parameters.

junchaowu:
will change handlerWrapper to a more meaningful name.

PR #68, opened 2016-03-21
Rename handlerWrapper to transportHandler

junchaowu:
This is to fix https://github.com/yarpc/yarpc-java/issues/46


yarpc-core/src/main/java/io/yarpc/handler/TransportHandler.java
```diff
@@ -27,8 +27,6 @@
 
 import java.nio.ByteBuffer;
 
-public interface HandlerWrapper {
-    Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest) throws RPCException;
-
+public interface TransportHandler extends Handler<ByteBuffer, ByteBuffer> {
     String getEncoding();
```
abhinav: You're not using the encoding anywhere according to the original ticket. Maybe delete it?

junchaowu: image
This part of code is commented out because our 4 language implementations are not on the same page yet. I believe in the future we will need this.

yarpc-core/src/main/java/io/yarpc/handler/TransportHandler.java
```diff
@@ -27,8 +27,6 @@
 
 import java.nio.ByteBuffer;
 
-public interface HandlerWrapper {
-    Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest) throws RPCException;
-
+public interface TransportHandler extends Handler<ByteBuffer, ByteBuffer> {
```
abhinav: This doesn't really address the problem you mentioned in person. You said you didn't want users to be able to do this (or discourage them from doing this):

    public class MyHandler implements Handler<ByteBuffer, ByteBuffer> { ... }
    rpc.register("foo", new MyHandler())

You said you want users to use handlers built via `{Raw,Json,Thrift}InboundHandler`. This interface doesn't solve that problem. I can still do that with one minor change:

    public class MyHandler implements TransportHandler { ... }
    rpc.register("foo", new MyHandler())

To get what you're trying to achieve, you'd have to do something like,

    public class TransportHandler {
        private Handler<ByteBuffer, ByteBuffer> handler
        // ...
    }

That is, make TransportHandler some sort of opaque container of wrapped handlers that you can get from JsonInboundHandler, etc.

abhinav:
I just realized that you're not using the name InboundHandler anywhere. You are using, `JsonInboundHandler`, `ThriftInboundHandler`, and `RawInboundHandler`. Do you think the class/interface should be called `InboundHandler` instead of `TransportHandler` or `HandlerWrapper`? It's slightly more specific than TransportHandler.

yarpc-core/src/main/java/io/yarpc/handler/TransportHandler.java
abhinav: Doesn't that stop me from writing a generic catch-all handler that just deals with bytes and doesn't care what encoding is being used?

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
```diff
@@ -34,7 +34,6 @@ public JsonInboundHandler(Handler handler) {
         super(new JsonSerializer(), handler);
     }
 
-    @Override
     public String getEncoding() {
         return Encoding.JSON;
     }
```
abhinav: Looks like this isn't used anymore?

yarpc-core/src/main/java/io/yarpc/handler/BaseInboundHandler.java
```diff
@@ -45,9 +45,7 @@ public BaseInboundHandler(Serializer serializer, Handler handler) {
         this.serializer = serializer;
     }
 
-    @Override
     public Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest) throws RPCException {
```
abhinav: Why'd you remove the `@Override`?

yarpc-core/src/main/java/io/yarpc/handler/InboundHandler.java
```diff
@@ -27,8 +27,6 @@
 
 import java.nio.ByteBuffer;
 
-public interface HandlerWrapper {
+public interface InboundHandler{
```
abhinav: Inconsistent spacing after the Handler. Please use your IDE to make sure you're formatting consistently across files.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
junchaowu: this is used in the unit test.

abhinav: But if it's not used in the code, what's the point of testing it?

junchaowu:
Format the code.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
junchaowu: Oh right.Removed the getEncoding method now.

abhinav:
:+1:

PR #69, opened 2016-03-24
Refactor context class

junchaowu:
- Add ttl and header into context class.
- Create builder for context class.
- Fix typo in previous build
- Fix project can't be built by Intellij issue
@yarpc/java


yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -43,8 +43,8 @@ public RPC(Config config) {
         this.config = config;
     }
 
-    public void register(String proceduer, InboundHandler handler) {
-        this.mHandlerRegistry.addHandler(proceduer, handler);
+    public void register(String procedure, InboundHandler handler) {
+        this.mHandlerRegistry.addHandler(procedure, handler);
```
abhinav: Not necessary to fix in this PR: Why is this called mHandlerRegistry? That's a very C++ pattern. In Java you should just use the camelCased name of the interface/class.

yarpc-core/src/main/java/io/yarpc/context/Context.java
```diff
@@ -21,7 +21,54 @@
...
+        return this.headers;
+    }
+
+    public void setHeaders(Map<String, String> headers) {
+        this.headers = headers;
+    }
+
+    public static class Builder {
+        private Duration nestedTtl;
+        private Map<String, String> nestedHeaders;
```
abhinav: Why are these attributes named `nested`? They can't conflict with the outer ones because this is a static nested class. When you do `this.ttl` anywhere in this Builder class, it can't possibly refer to `Context.ttl` because a context instance doesn't exist in this scope.

abhinav:
:ok:

PR #70, opened 2016-03-28
feat: introduce 'Channel' class

junchaowu:
Implement 'Channel' class
Make the 'Client' to accept 'Channel' instance in the constructor


junchaowu:
@yarpc/yarpc  This is to make usage of client be more like 'yarpc-go' style with 'Channel' concept.

yarpc-core/src/main/java/io/yarpc/Channel.java
```diff
@@ -0,0 +1,89 @@
...
+ * THE SOFTWARE.
+ */
+package io.yarpc;
+
+import io.yarpc.transport.Outbound;
+
+public class Channel {
+    private String service;
+    private String caller;
+    private Outbound outbound;
```
abhinav: As with the Go design: Add a TODO here to add per-channel timeouts, etc.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -96,4 +98,13 @@ public void stop() {
...
         this.runningFuture.set(null);
     }
+
+    public Channel createChannel(String service) throws OutboundNotFoundException {
+        return this.createChannel(service, this.name);
+    }
+
+    public Channel createChannel(String service, String caller) throws OutboundNotFoundException {
+        Outbound outbound = this.lookUpOutbound(service);
+        return new Channel.Builder().service(service).caller(caller).outbound(outbound).build();
```
abhinav: IMO caller shouldn't be customizable by the user. Caller is always the name specified when the RPC was constructed. 

(If they want to, they can still manually `setCaller` on the Channel object.)

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -38,21 +38,24 @@
...
-        Outbound outbound = this.rpc.lookUpOutbound(request.getService());
+
         request.setEncoding(Encoding.JSON);
         logger.debug(String.format("sending request %s", request.toString()));
 
         Request<ByteBuffer> transportRequest = Request.toByteBufferRequest(request, JsonClient.serializer);
         RPCFuture<Response<O>> rpcFuture = new RPCFuture<>();
+        Outbound outbound = this.channel.getOutbound();
+        transportRequest.setCaller(this.channel.getCaller());
+        transportRequest.setService(this.channel.getService());
```
abhinav: You want to `set{Caller, Service}` only if their current values are null. If the user's request had its own values, then maybe we should keep those.

yarpc-core/src/main/java/io/yarpc/RPC.java
breerly: +1

yarpc-core/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
```diff
@@ -39,22 +39,23 @@
...
             responseBodyType) throws Exception {
-        Outbound outbound = this.rpc.lookUpOutbound(request.getService());
-
         request.setEncoding(Encoding.THRIFT);
         logger.debug(String.format("sending request %s", request.toString()));
 
         Request<ByteBuffer> transportRequest = Request.toByteBufferRequest(request, ThriftClient.serializer);
         RPCFuture<Response<O>> rpcFuture = new RPCFuture<Response<O>>();
+        Outbound outbound = this.channel.getOutbound();
+        transportRequest.setCaller(this.channel.getCaller());
```
breerly: nice

yarpc-core/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -126,12 +125,10 @@ public void tchannelTest() throws Exception {
     }
 
     private void call(String server) throws Exception {
-        JsonClient client = new JsonClient(rpc);
+        JsonClient client = new JsonClient(rpc.createChannel(server));
```
breerly: any reason we shouldn't just go with `rpc.channel(server)` here like go does?

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
```diff
@@ -128,12 +127,10 @@ public void tchannelTest() throws Exception {
     }
 
     private void call(String server) throws Exception {
-        RawClient client = new RawClient(rpc);
+        RawClient client = new RawClient(rpc.createChannel(server));
```
breerly: I think a cleaner API is to use `rpc.channel(server)` since this will be called alot

breerly:
Looks good. I had a naming nit, but that might not be idiomatic. 

:shipit:

yarpc-core/src/test/java/io/yarpc/encoding/raw/RawIntegrationTest.java
abhinav: I'm okay with channel or createChannel. FWIW this won't be called a lot because you will usually construct the client for a specific service once and then use that for all requests to that service.

abhinav:
LGTM overall. Minor comments.

PR #73, opened 2016-03-28
refactor: Move common parts from Json/Raw/ThriftClient to RPCClient

junchaowu:
Most of the sending and receiving logic in RPCClient are the same across
three encoding types. So move common piece of work into abstract base class.


junchaowu:
@yarpc/yarpc

junchaowu:
ping @yarpc/yarpc

breerly:
:ship:

abhinav:
:+1:

PR #74, opened 2016-03-29
Update build config file.

junchaowu:
Include compile dependency into yarpc-core jar file.
Add help target for Makefile
@yarpc/java


breerly:
:shipit:

PR #75, opened 2016-04-07
feat: create Exception Message properties

junchaowu:
Before code review for major changes for errors, I first submit small part of it to show I manage the exception message template.


junchaowu:
@yarpc/yarpc

breerly:
Sorry for the delay, I'll get eyes on this tomorrow!

yarpc-core/src/main/resources/ExceptionMessage.properties
```diff
@@ -0,0 +1,10 @@
+INVALID_TTL =  Bad Request: invalid TTL "%s" for procudure "%s" of service "%s": must be positive integer\n
```
kriskowal: \n should not be a part of the error message. the \n should be added by the HTTP transport to the message body. the \n should not exist in the tchannel error frame message

yarpc-core/src/main/java/io/yarpc/exception/ExceptionMessage.java
```diff
@@ -0,0 +1,38 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.exception;
+
+import java.util.ResourceBundle;
```
kriskowal: perhaps this class provides an unnecessary level of indirection. you could load the resource bundle and call rb.getString at each exception site.

yarpc-core/src/main/resources/ExceptionMessage.properties
```diff
@@ -0,0 +1,10 @@
...
+FAIL_ENCODE_HEADER = failed to encode "%s" request headers for procedure "%s" of service "%s": %s\n
+FAIL_DECONDE_REQ_HEADER = BadRequest: failed to decode "%s" request headers for procedure "%s" of service "%s": %s\n
```
kriskowal: i recommend spelling out REQ and RESP as REQUEST and RESPONSE. In JavaScript, I’ve been using req and res per our local idiom, for transport request and response, but spelling out the RPC request and response.

yarpc-core/src/main/resources/ExceptionMessage.properties
kriskowal: no space between Bad and Request

yarpc-core/src/main/java/io/yarpc/exception/ExceptionMessage.java
abhinav: A small merit to this approach is that if any of the names used here are undefined, it will fail right away at start up rather than when the `getString` call is made.

yarpc-core/src/main/java/io/yarpc/exception/ExceptionMessage.java
```diff
@@ -0,0 +1,38 @@
...
+
+public final class ExceptionMessage {
+    public static ResourceBundle rb = ResourceBundle.getBundle("ExceptionMessage");
```
abhinav: Make `rb` private if the intention is for actual usage of the values to always go through the constants you have declared.

kriskowal:
This is a good start. Let’s call it a WIP until it’s bundled up with tests.

yarpc-core/src/main/java/io/yarpc/exception/ExceptionMessage.java
```diff
@@ -0,0 +1,38 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.exception;
+
+import java.util.ResourceBundle;
+
+public final class ExceptionMessage {
```
abhinav: Consider using a plural in the name: `ExceptionMessages` or `ErrorMessages`

yarpc-core/src/main/resources/ExceptionMessage.properties
abhinav: Inconsistency: `BadRequest` vs `Bad Request`.

yarpc-core/src/main/resources/ExceptionMessage.properties
abhinav: +1. Don't make newlines part of the message

yarpc-core/src/test/java/io/yarpc/RPCTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class RPCTest {
+    @Test
+    public void testOutBoundNotFound() throws IOException, InterruptedException {
```
abhinav: Outbound, not OutBound

abhinav:
One thing that really bothers me about this approach is that it's very easy for the format strings to get out of sync with the arguments they use and there's no check for that. I've used MessageFormat and resource files similarly in the past, and it's really annoying to try to debug something only to find a `%s` instead of the value you expect there because you forgot to add the extra argument to `String.format`.

breerly:
@abhinav what do you suggest then?

abhinav:
Options:

* Deal with it: Keep this as-is and just try not to get the format string and the arguments out of sync. This works if the number of format strings we're putting in the file remains small.
* A function to go with each format string rather than an entry in the file. I'm not a big fan of this option. It adds too much boilerplate.
* A unique exception type for each case (that inherits `BadRequestException`/`UnexpectedException`/what-have-you), which accepts the information necessary for formatting in the constructor instead of accepting the formatted message. Then the class itself needs to know how to format the message for the error it represents. So, `new OutboundNotFoundException(service)`, not `new OutboundNotFoundException(String.format(...))`. I prefer this approach because,
  * It doesn't add much boilerplate beyond what we're already doing.
  * All formatting of a specific error will happen in the same place rather than at all `throw` sites. That is, instead of `new OutboundNotFoundException(String.format(...))` in 5 places, you have `new OutboundNotFoundException(service)`, in 5 places, and `String.format(..., service)`, in a single place. This is especially useful if we decide to add another parameter.
  * It is possible for the exception class to store the parameters as attributes and provide runtime access to them when the exceptions are caught.

          try {
              ...
          } catch (OutboundNotFoundException e) {
              if (e.getService() == "something") { ... }
          }

breerly:
The third option seems ideal - @junchaowu @kriskowal you guys onboard?

junchaowu:
After discussion offline, we will keep the approach now, if in the future we are able to pass the exception type information through wire, we may switch to different approach. Right now we just keep the big category exception type.

breerly:
:shipit:

PR #77, opened 2016-04-08
Add Yarpc error handling for both http and tchannel

junchaowu:
@yarpc/java


blampe:
This is a big diff, is it possible to break it up?

junchaowu:
change it to branch crossdock-errors to match  crossdock branch name.

PR #82, opened 2016-04-13
HTTP/TChannel Server side error handling

junchaowu:
Test will be added in the future PRs. The client side test changes will be in a different PR.
@yarpc/java


yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
```diff
@@ -33,4 +34,8 @@ public JsonInboundHandler(Handler handler) {
         super(new JsonSerializer(), handler);
     }
 
+    @Override
+    public String getEncoding() {
```
junchaowu: the Encoding is added for the purpose to let the incoming request having encoding type. It will be used to compose the error message.

docker-compose.yml
```diff
@@ -25,7 +25,7 @@ services:
             - "8080-8082"
 
     node:
-        image: yarpc/yarpc-node
+        build: ../yarpc-node
```
breerly: this will break your build, you'll want to switch it back to `image:`

junchaowu:
the test coverage is decreased because I haven't pushed tests for errors. I plan to send out tests for code review after all client and server side error handling ready.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
breerly: I could buy that.

breerly:
This is a good start :shipit:

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
```diff
@@ -33,4 +34,8 @@ public JsonInboundHandler(Handler handler) {
...
+        return Encoding.JSON;
```
abhinav: Probably for a different PR: I recommend deleting the global Encoding class in favor of each encoding implementation knowing what its name is and making it available from constants defined in their respective packages. Depending on a central list of encodings defeats one of the design goals of YARPC: The ability to define new encodings without modifying the core library.

breerly: Good catch :+1:

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
```diff
@@ -46,13 +49,51 @@ public AbstractInboundHandler(Serializer serializer, Handler handler) {
     }
 
     @Override
-    public Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest) throws RPCException {
+    @SuppressWarnings("unchecked")
```
abhinav: Where in this function are you doing the unchecked cast specifically? It's better to have the annotation above the variable assignment rather than on the whole function because it hides what could be a legitimate problem.

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
```diff
@@ -46,13 +49,51 @@ public AbstractInboundHandler(Serializer serializer, Handler handler) {
...
+    public Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest)
+            throws InvalidInboundRequestException, UnexpectedException, InvalidOutboundResponseException {
+        Request request;
+        try {
+            request = Request.fromByteBufferRequest(transportRequest, this.serializer, requestBodyType);
+        } catch (Exception e) {
+            throw new InvalidInboundRequestException(
```
abhinav: Use the `Exception(message, cause)` constructor so that we have the full stack trace.

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
```diff
@@ -46,13 +49,51 @@ public AbstractInboundHandler(Serializer serializer, Handler handler) {
...
+            response = this.getHandler().handle(request);
+        } catch (Exception e) {
+            throw new UnexpectedException(String.format(
+                    ExceptionMessages.UNEXPECTED,
+                    request.getProcedure(),
+                    request.getService(),
+                    e.getMessage()));
+        }
+
+        if (response == null) {
```
abhinav: I don't think you need to explicitly handle this here simply because returning a non-null result is part of the contract of `handle()`. We're okay with crashing if we have that contract violation. The layer above this can handle this failure. Also, the `Response.toByteBufferResponse` below will also fail if `response` is null so it's okay to let that handle it.

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.common.base.Strings;
+import humanize.Humanize;
+import io.yarpc.exception.ExceptionMessages;
+import io.yarpc.exception.InvalidInboundRequestException;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public abstract class RequestParser {
+    public static void checkRequireProperties(
```
abhinav: `checkRequiredProperties`

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
```diff
@@ -0,0 +1,61 @@
...
+            String service,
+            String procedure,
+            String caller,
+            String ttl) throws
```
abhinav: Consider creating a parameter object (you can nest it inside this class) instead. Getting the arguments for this function out of order will be easy given that the types are all the same.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPRequestParser.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class HTTPRequestParser extends RequestParser {
+    public static Request<ByteBuffer> parse(FullHttpRequest httpRequest) throws InvalidInboundRequestException {
+        HttpHeaders httpHeaders = httpRequest.headers();
+        String service = httpHeaders.get(Names.SERVICE);
+        String procedure = httpHeaders.get(Names.PROCEDURE);
+        String caller = httpHeaders.get(Names.CALLER);
+        String ttlMsS = httpHeaders.get(Names.TTLMS);
+
+        checkRequireProperties(service, procedure, caller, ttlMsS);
```
abhinav: Inheritance is not the way to do this. The method is already public static. Just use it instead of inheriting the class.

yarpc-core/src/main/java/io/yarpc/handler/InboundHandler.java
```diff
@@ -29,4 +29,6 @@
 
 public interface InboundHandler {
     Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest) throws RPCException;
+
+    String getEncoding();
```
abhinav: I recommend against doing this. This makes it impossible to implement InboundHandlers that are backed by other handlers. The Registry can implement the InboundHandler interface very safely if you drop this requirement.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -62,48 +65,49 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
     }
 
-    private Request<ByteBuffer> parseRequest(FullHttpRequest httpRequest) {
-        Map<String, String> headers = new HashMap<>();
-        for (Map.Entry<String, String> entry : httpRequest.headers().entries()) {
-            headers.put(entry.getKey(), entry.getValue());
+    private void forwardRequest(ChannelHandlerContext ctx, FullHttpRequest httpRequest) throws Exception {
+        try {
+            Request<ByteBuffer> transportRequest = HTTPRequestParser.parse(httpRequest);
+            InboundHandler inboundHandler = this.handlerRegistry.lookupHandler(transportRequest.getProcedure());
```
abhinav: For a different diff: The HTTP/TChannel handler should not know about the registry. It should consume a InboundHandler. The Registry should implement the InboundHandler interface (which means you'll have to drop the `getEncoding` on the handler).

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -62,48 +65,49 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
+            if (inboundHandler == null) {
+                throw new InvalidInboundRequestException(
+                        String.format(
+                                ExceptionMessages.UNRECOGNIZED_HANDLER,
+                                transportRequest.getProcedure(),
+                                transportRequest.getService()));
+            }
+            transportRequest.setEncoding(inboundHandler.getEncoding());
```
abhinav: The encoding is included in the YARPC headers. You should be using that instead of what the handler thinks it should be.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -62,48 +65,49 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
+            Response<ByteBuffer> transportResponse = inboundHandler.handle(transportRequest);
+            sendResponse(ctx, transportResponse, transportRequest);
+
```
abhinav: Nit: Delete empty line

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -62,48 +65,49 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
-        // TODO build yarpc common context including tracing, meta info.
-        // Context context = ContextBuilder.build(httpRequest);
-
-        Request<ByteBuffer> transportRequest = this.parseRequest(httpRequest);
-        Response<ByteBuffer> transportResponse = inboundHandler.handle(transportRequest);
-
+    private void sendResponse(
+            ChannelHandlerContext ctx, Response<ByteBuffer> transportResponse, Request<ByteBuffer> transportRequest) {
+        ByteBuffer body = transportResponse.getBody();
+        ByteBuf bytebuf ;
```
abhinav: what's with the space before the semicolon

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -62,48 +65,49 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
+        if (transportRequest.getEncoding().equals(Encoding.JSON)) {
+            bytebuf = Unpooled.wrappedBuffer(2, body, ByteBuffer.wrap(NEWLINE));
+        } else {
+            bytebuf = Unpooled.wrappedBuffer(body);
+        }
```
abhinav: What's the purpose of this? This doesn't look like something the transport should know or care about.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -112,30 +116,49 @@ private void forwardRequest(ChannelHandlerContext channelHandlerContext, FullHtt
...
-        return inboundHandler;
+    private void sendError(ChannelHandlerContext ctx, ServerException e, HttpResponseStatus status) {
+        FullHttpResponse httpResponse = new DefaultFullHttpResponse(
+                HttpVersion.HTTP_1_1,
+                status,
+                Unpooled.wrappedBuffer(
+                        e.getMessage().getBytes(),
+                        // Add "\n" for HTTP protocol only
+                        NEWLINE
+                )
```
abhinav: ?

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelRequestParser.java
```diff
@@ -0,0 +1,111 @@
...
+
+        String caller = null;
+        if (tchannelRequest.getTransportHeaders().containsKey(TransportHeaders.CALLER_NAME_KEY)) {
+            caller = tchannelRequest.getTransportHeaders().get(TransportHeaders.CALLER_NAME_KEY);
+        }
+
+        Duration timeout = Duration.millis(tchannelRequest.getTimeout());
+        checkRequireProperties(service, procedure, caller, String.valueOf(tchannelRequest.getTimeout()));
+
+        if (timeout.getMillis() < 0) {
```
abhinav: `<= 0`

0 is not a valid timeout either

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelRequestParser.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private static String translateEncoding(ArgScheme argScheme) throws InvalidInboundRequestException {
+        switch (argScheme) {
+            case JSON:
+                return Encoding.JSON;
+            case RAW:
+                return Encoding.RAW;
+            case THRIFT:
+                return Encoding.THRIFT;
```
abhinav: just use `argScheme.getScheme()` to get its name. Don't hard code the list of supported schemes in this layer

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelServerInboundHandler.java
```diff
@@ -0,0 +1,140 @@
...
+    private final static Logger logger = LoggerFactory.getLogger(TChannelServerInboundHandler.class);
+    private final static AttributeKey<Long> REQUEST_ID = AttributeKey.valueOf("request_id");
+    private final HandlerRegistry handlerRegistry;
+
+    public TChannelServerInboundHandler(HandlerRegistry handlerRegistry) {
+        this.handlerRegistry = handlerRegistry;
+    }
+
+    public static ByteBuf encodeHeaderForTChannel(io.yarpc.Request<ByteBuffer> transportRequest, io.yarpc
+            .Response<ByteBuffer> transportResponse) throws InvalidOutboundResponseException {
```
abhinav: Nit: pull the whole type for the second parameter down rather than terminating the line half way through

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelServerInboundHandler.java
```diff
@@ -0,0 +1,140 @@
...
+        ByteBuf headers = null;
+        try {
+            if (transportRequest.getEncoding().equals(Encoding.JSON)) {
+                headers = HeaderUtil.encodeHeadersJsonStyle(transportResponse.getHeaders());
+            } else {
+                headers = HeaderUtil.encodeHeaders(transportResponse.getHeaders());
+            }
+            return headers;
+        } catch (Exception e) {
+            throw new InvalidOutboundResponseException(
```
abhinav: What's with these catch-all exceptions everywhere? `encodeHeaders` raises nothing and `encodeJsonStyle` raises only `JsonProcessingException`. Be specific about the exceptions you're catching where possible. Encoding and decoding should have a very narrow set of possible exceptions.

yarpc-core/src/test/java/io/yarpc/transport/tchannel/TChannelRequestParserTest.java
```diff
@@ -0,0 +1,133 @@
...
+        } catch (InvalidInboundRequestException e) {
+            Assert.assertEquals(e.getMessage(), String.format(ExceptionMessages.MISSING, "service name"));
+        }
+    }
+
+    @Test
+    public void testMissingProcedure() throws IOException {
+        Request request = Mockito.mock(Request.class);
+        Mockito.when(request.getService()).thenReturn("s");
+        Mockito.when(request.getEndpoint()).thenReturn(null);
```
abhinav: Request is a POJO. Instead of mocking it, just build one by hand.

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
junchaowu: I disagree. The string type is String which mean if user has custom handler, they can return the custom string. It doesn't necessary to be Yarpc-java's encoding. The global Encoding class I have so at least I can be consistent for the encoding type inside the yarpc since it is a String.

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
junchaowu: The whole idea for my error handling if to capture the error at specific places since we have specific error message for different error cases. If I skip the null here it may be captured by other error check but it will mislead users or clients since the root cause is not there.

yarpc-core/src/main/java/io/yarpc/handler/InboundHandler.java
junchaowu: If I drop this, how do you know the encoding type for an incoming request and the encoding type will be used in the error message composition.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPRequestParser.java
junchaowu: I also don't quite like it. I will put it in a common places.

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
junchaowu: In java I can't have annotation on arbitrary expressions. 
the unchecked one is for 
```
@SuppressWarnings("unchecked") /// ok
Response response = this.getHandler().handle(request);
```
However in my code in order to wrap the line with try catch exception. Right now java compile compliant about the uncheck annotation.
```
Response response;
try{
    @SuppressWarnings("unchecked") /// not ok
    response = ...
}
```

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
junchaowu: good suggestion.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
junchaowu: We haven't synced on the headers or contexts yet. If that's true, in another task, I will modify it.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
junchaowu: @kriskowal may have better explanation on this one.  This new line sign is used for http json response. when people curl the http json endpoint, in order to make the response more user friendly, we add this.

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelRequestParser.java
junchaowu: will we use 0 as no timeout in the future?

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelRequestParser.java
junchaowu: ``argScheme.getScheme() `` will bind us tightly with tchannel. Because it returns a string. I am afraid the string return by tchannel isn't exact same as what yarpc has.

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelServerInboundHandler.java
junchaowu: That's true

yarpc-core/src/main/java/io/yarpc/encoding/json/JsonInboundHandler.java
abhinav: The intention is to avoid having an explicit list of encodings anywhere. I understand the motivation for having a constant that holds the name, and that's generally a good idea, but instead of putting them all in the same class, maybe put them in the encoding implementations inside the packages implementing the encodings.

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
abhinav: Yeah, you can't do arbitrary expressions. I was hoping you could work around the try catch restriction, possibly by adding helper methods. Generally `@SuppressWarnings` should be as narrow as possible for the sake of safety.

abhinav: Also, for a different PR:

`Request.fromByteBufferRequest` is,

        public static Request fromByteBufferRequest(Request<ByteBuffer> request, Serializer serializer, Class<?> bodyType)

Unless I'm misunderstanding this, it can safely be,

    public static <T> Request<T> fromByteBufferRequest(Request<ByteBuffer> request, Serializer serializer, Class<T> bodyType)

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
abhinav: But it won't mislead users. The other exception will include the root cause and the message from it.

There's a difference between errors which we expect to occur and handle, and contract violations, which are bugs. Getting a ull Response from `handle` is a bug, not something we expect to happen ever in production. So it doesn't need to be handled like this. It should just cause a NullPointerException and fail.

yarpc-core/src/main/java/io/yarpc/handler/InboundHandler.java
abhinav: Shouldn't that be known by the encoder? If I'm trying to parse a JSON payload, I know that the name of the encoding is 'json' and I include that in the error.

What I'm trying to get at here is this: Handler takes a `Request<ByteBuffer>` and returns a `Response<ByteBuffer>`. With such a simple interface, you can have `Registry` implement this interface and dispatch on `request.getProcedure()` against registered procedures. You can have `RPC` also implement the interface and simply delegate to the `Registry` it holds. An `Inbound` doesn't need to know about the `Registry`, just a Handler. It builds a `Request<ByteBuffer>` from an incoming request and passes it off to the Handler which takes care of dispatch, decoding, etc. at different layers.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
abhinav: Synced on headers? The encoding header has been part of the properties doc for a while: https://github.com/yarpc/yarpc/blob/master/properties.md#encoding. We only need to sync on whether it's required or optional (I think other languages treat it as optional right now).

Regardless, `InboundHandler` should not know about the encoding. The inbound `Request` can have an unset encoding if the header is missing. You don't need to get the encoding from the `Request` to know what kind of encoding/decoding error you ran into. That is known by the encoding implementation.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
abhinav: The RPC protocol shouldn't have a JSON-specific hack for the `curl` use case. That's not the primary mode of usage. Good shell prompts account for command output not ending with an empty line. We don't need a hack for that.

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
abhinav: ping

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
abhinav: ping

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelRequestParser.java
abhinav: Not that I know of. 0 is an invalid value for a timeout. And although we haven't got to that part of the discussion yet, "no timeout" is probably invalid usage of the library.

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelRequestParser.java
abhinav: Okay. I can buy the need for hard coding the list because the TChannel transport only supports these three encodings for now. We need to fix the TChannel library to not use that enum and use a string or a custom non-enum type instead if we want this to work with other encodings.

yarpc-core/src/test/java/io/yarpc/transport/tchannel/TChannelRequestParserTest.java
abhinav: Ping

abhinav:
Second pass completed. Left comments/replied.

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
```diff
@@ -0,0 +1,61 @@
...
+            missings.add("caller name");
+        }
+
+        if (Strings.isNullOrEmpty(ttl)) {
+            missings.add("TTL");
+        }
+
+        if (missings.size() > 0) {
+            throw new InvalidInboundRequestException(
+                    String.format(ExceptionMessages.MISSING, Humanize.oxford(missings)));
```
abhinav: I don't think the extra dependency on humanize is worth this. Doing it by hand isn't that difficult.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -112,30 +115,49 @@ private void forwardRequest(ChannelHandlerContext channelHandlerContext, FullHtt
...
-        // look up user procedure.
-        String procedure = rawRequest.headers().get(Names.PROCEDURE).toString();
-        InboundHandler inboundHandler = this.handlerRegistry.lookupHandler(procedure);
-        if (inboundHandler == null) {
-            throw new RPCException(String.format("Can't find handler for procedure: %s.", procedure));
+    private HttpResponseStatus getStatusCode(ServerException e) {
+        if (e instanceof InvalidInboundRequestException) {
+            return HttpResponseStatus.BAD_REQUEST;
+        } else {
+            return HttpResponseStatus.INTERNAL_SERVER_ERROR;
```
abhinav: Might be worth adding a `getHttpStatusCode` method to the `ServerException` class and overriding it in `InvalidInboundRequestException`? It's not a perfect solution because now the exception type knows about HTTP, so I'm not 100% sold on it either.

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelRequestParser.java
```diff
@@ -0,0 +1,111 @@
...
+
+public class TChannelRequestParser {
+    public static io.yarpc.Request<ByteBuffer> parse(Request tchannelRequest) throws InvalidInboundRequestException {
+        String service = tchannelRequest.getService();
+
+        String procedure = tchannelRequest.getEndpoint();
+
+        String caller = null;
+        if (tchannelRequest.getTransportHeaders().containsKey(TransportHeaders.CALLER_NAME_KEY)) {
+            caller = tchannelRequest.getTransportHeaders().get(TransportHeaders.CALLER_NAME_KEY);
```
abhinav: Put `tchannelRequest.getTransportHeaders()` into a local variable

abhinav:
Update: Second pass completed. Left comments/replied.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
kriskowal: @abhinav I disagree. Customers will be disappointed and the customer is always right. Perhaps we can make this less hacky by moving the \n into the encoding implementation.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
junchaowu: That's my original plan. But that's will bind us to the protocol. So I choose not to do that.

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
junchaowu: changed the fromByteBufferRequest to generics

yarpc-core/src/test/java/io/yarpc/transport/tchannel/TChannelRequestParserTest.java
junchaowu: Both are okay. Won't fix in this PR

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
abhinav: I agree it's not a perfect solution, but IMO it's better than `instanceof` checks.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
abhinav: I can live with the encoding implementation adding the newline. Created https://github.com/yarpc/yarpc-java/issues/86

FWIW Go's [JSON encoder](https://golang.org/pkg/encoding/json/#Encoder.Encode) always adds a newline character when using that specific API.

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
junchaowu: I won't bother this since we can build it for computers not resource limit. Node is also using a third party lib. Shouldn't a problem

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
junchaowu: I don't see problem's with instanceof. It won't hurt performance.

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
```diff
@@ -46,13 +49,52 @@ public AbstractInboundHandler(Serializer serializer, Handler handler) {
     }
 
     @Override
-    public Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest) throws RPCException {
+    @SuppressWarnings("unchecked")
+    public Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest)
+            throws InvalidInboundRequestException, UnexpectedException, InvalidOutboundResponseException {
+        transportRequest.setEncoding(serializer.getEncoding());
+        Request request;
```
abhinav: We can fix this later:

- Always provide arguments for generics. If they're not known, use `<?>`.
- Always use `this.` when accessing instance variables.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPRequestParser.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Map;
+
+public class HTTPRequestParser {
+    public static Request<ByteBuffer> parse(FullHttpRequest httpRequest) throws InvalidInboundRequestException {
+        HttpHeaders httpHeaders = httpRequest.headers();
+        RequestParser.RequiredProperty requiredProperty = new RequestParser.RequiredProperty();
+        requiredProperty.setService(httpHeaders.get(Names.SERVICE));
+        requiredProperty.setProcedure(httpHeaders.get(Names.PROCEDURE));
+        requiredProperty.setCaller(httpHeaders.get(Names.CALLER));
+        requiredProperty.setTtl(httpHeaders.get(Names.TTLMS));
```
abhinav: For later (consider creating a GH issue): there's inconsistency here in how abbreviations are handled. You're using `HTTP` in some places and `Http` in others, and similarly, you're using `Ttl` and `TTL` mixed (some tests are using TTL).

junchaowu: https://github.com/yarpc/yarpc-java/issues/87

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
abhinav: It appears that you don't actually need the `instanceof`. Above, you're doing,

            sendResponse(ctx, transportResponse, transportRequest);
        } catch (ServerException e) {
            sendError(ctx, e, getStatusCode(e));

Instead, do,

            sendResponse(ctx, transportResponse, transportRequest);
        } catch (InvalidInboundRequestException e) {
            sendError(ctx, e, HttpResponseStatus.BAD_REQUEST);
        } catch (ServerException e) {
            sendError(ctx, e, HttpResponseStatus.INTERNAL_SERVER_ERROR);

And below, you're doing `getStatusCode` with something known to be `UnexpectedException`, so you can just use `HttpResponseStatus.INTERNAL_SERVER_ERROR`.

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelServerInboundHandler.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private final static Logger logger = LoggerFactory.getLogger(TChannelServerInboundHandler.class);
+    private final static AttributeKey<Long> REQUEST_ID = AttributeKey.valueOf("request_id");
+    private final InboundHandler dispatchHandler;
+
+    public TChannelServerInboundHandler(InboundHandler dispatchHandler) {
+        this.dispatchHandler = dispatchHandler;
+    }
+
+    public static ByteBuf encodeHeaderForTChannel(
```
abhinav: why is this public

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelServerInboundHandler.java
```diff
@@ -0,0 +1,135 @@
...
+            // to tchannel headers
+            ByteBuf headers = encodeHeaderForTChannel(transportRequest, transportResponse);
+
+            RawResponse response = new RawResponse.Builder(request)
+                    .setArg2(headers)
+                    .setArg3(Unpooled.wrappedBuffer(transportResponse.getBody()))
+                    .build();
+            sendResponse(ctx, response);
+        } catch (ServerException e) {
+            ErrorFrame.sendError(toTChannelErrorType(e), e.getMessage(), request.getId(), ctx);
```
abhinav: Same as HTTP, catch the specific exceptions rather than catching generic and then doing an instanceof

abhinav:
Down to the last of the issues. Should be good to go soon.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
junchaowu: done

yarpc-core/src/main/java/io/yarpc/transport/tchannel/TChannelServerInboundHandler.java
junchaowu: changed to private

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
```diff
@@ -62,48 +66,41 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
-            headers.put(entry.getKey(), entry.getValue());
+    private void forwardRequest(ChannelHandlerContext ctx, FullHttpRequest httpRequest) throws Exception {
+        try {
+            Request<ByteBuffer> transportRequest = HTTPRequestParser.parse(httpRequest);
+            Response<ByteBuffer> transportResponse = this.dispatchHandler.handle(transportRequest);
+            sendResponse(ctx, transportResponse, transportRequest);
+        } catch (InvalidInboundRequestException e) {
+            sendError(ctx, e, HttpResponseStatus.BAD_REQUEST);
+        } catch (InvalidOutboundResponseException e) {
+            sendError(ctx, e, HttpResponseStatus.INTERNAL_SERVER_ERROR);
```
abhinav: Is InvaildOutboundResponseException the only exception that can be raised here (besides unexpected failures)?

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
abhinav: That's not the point of my comment, though. It's okay for us to use external dependencies, but as a library anything you depend on is now a transitive dependency for consumers. So you have to be more selective about what you want to add to the dependency tree. You're adding a dependency to call a single function that is only slightly more complicated than calling `String.join`. It's trivial to implement it by hand. There is no need for this dependency.

abhinav: To clarify: This doesn't need to be done in this PR. It can wait for another PR.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPServerInboundHandler.java
junchaowu: The invalidInboundRequestException is above it.

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
abhinav: Created https://github.com/yarpc/yarpc-java/issues/88

abhinav:
👍

PR #84, opened 2016-04-14
Use RPC prefix and Context-TTL-MS

breerly:
* use `RPC-` prefix instead of `YARPC-` prefix.
* use `Context-TTL-MS` instead of `YARPC-Ttlms`.


junchaowu:
thanks for changing this.

PR #85, opened 2016-04-18
Add inboundhandler interface to the handlerRegistry

junchaowu:


yarpc-core/src/main/java/io/yarpc/MapHandlerRegistry.java
```diff
@@ -45,4 +50,18 @@ public InboundHandler lookupHandler(String procedure) {
         }
         return this.handlers.get(procedure);
     }
+
+    @Override
+    public Response<ByteBuffer> handle(Request<ByteBuffer> transportRequest) throws RPCException {
+        InboundHandler inboundHandler = this.lookupHandler(transportRequest.getProcedure());
+        if (inboundHandler == null) {
+            throw new InboundHandlerNotFoundException(
```
abhinav: Shouldn't this be using `InvalidInboundRequestException`? The same thing that was being raised previously? Or if you're switching to this exception, is the other one not used anymore?

yarpc-core/src/main/java/io/yarpc/MapHandlerRegistry.java
```diff
@@ -22,8 +22,13 @@
 package io.yarpc;
 
 
+import io.yarpc.exception.ExceptionMessages;
+import io.yarpc.exception.InboundHandlerNotFoundException;
+import io.yarpc.exception.InvalidInboundRequestException;
+import io.yarpc.exception.RPCException;
```
abhinav: Unused import?

abhinav:
LGTM

PR #89, opened 2016-04-20
Crossdock errors testing

junchaowu:


docker-compose.yml
```diff
@@ -5,44 +5,46 @@ services:
         links:
             - java
             - node
-            - go
-            - python
-            - python-sync
+#            - go
+#            - python
+#            - python-sync
```
breerly: Can you uncomment these? Behaviors that aren't implemented should just be skipped, they shouldn't fail the test suite.

docker-compose.yml
```diff
@@ -5,44 +5,46 @@ services:
...
             - AXIS_TRANSPORT=http,tchannel
-
-            - BEHAVIOR_RAW=client,server,transport
-            - BEHAVIOR_JSON=client,server,transport
-            - BEHAVIOR_THRIFT=client,server,transport
+            - AXIS_ERRORS=node
+#
+#            - BEHAVIOR_RAW=client,server,transport
+#            - BEHAVIOR_JSON=client,server,transport
+#            - BEHAVIOR_THRIFT=client,server,transport
```
breerly: lets make sure these are uncommented before landing.

This feature will be useful for keeping these uncommented: https://github.com/yarpc/crossdock/issues/44

docker-compose.yml
junchaowu: can't do it right. The go and python will fail on crossdock test

abhinav:
Just an update: You can test this against `yarpc-{go,node}:crossdock-errors` now. You have a few minor things to fix:

The error you raise for `unexpected-error` is wrong.

- Go

  ```
  Error: Not equal: "UnexpectedError: error for procedure \"unexpected-error\" of service \"yarpc-test\": error\n" (expected)
          != "UnexpectedError: error for procedure \"unexpected-error\" of service \"yarpc-test\": bad boy\n" (actual)
  Messages: unexpected error: response body should be informative error
  ```

- Node

  ```
  expected: 'UnexpectedError: error for procedure "unexpected-error" of service "yarpc-test": error\n'
  actual: 'UnexpectedError: error for procedure "unexpected-error" of service "yarpc-test": bad boy\n'
  ```

Also, calls from Java to Node, Go, and Java for the `errors` behavior simply
fail:

```
F - {0xc820013020 java map[behavior:errors server:go]} - err: net/http: request canceled
F - {0xc820013020 java map[behavior:errors server:node]} - err: net/http: request canceled
F - {0xc820013020 java map[behavior:errors server:java]} - err: net/http: request canceled
```

junchaowu:
The 'bad body' is the actually error message I throw in the function. Java and node test pass before they should only check the template part.

junchaowu:
errors behaviors java is the server , only node will be the client. because node implements a bad node client which will send out bad requests . We should only use node as client for error behaviors.

abhinav:
Whether we use just Node or Node and Go for the test client is not really relevant to this discrepancy. The behavior of the test subjects (the RPC server implementations) has to be the same in all languages. yarpc-node's implementation of the 'unexpected-error' procedure on the test subject is here: https://github.com/yarpc/yarpc-node/blob/crossdock-errors/test/subject.js#L188

```javascript
    this.json.register({
        procedure: 'unexpected-error'
    }, function error(request, callback) {
        callback(new Error('error'));
    });
```

So your implementation of the test subject should return the same error and message. We'll want to update the [test-subject.md](https://github.com/yarpc/yarpc/blob/master/test-subject.md) with this as well: The test subject must implement an `unexpected-error` procedure that fails with an unexpected error with the message "error".

junchaowu:
sure l'll update that.

junchaowu:
this PR is deprecated because the middle change will have latest error handling.

PR #90, opened 2016-04-21
HTTP: Prepend application headers with `Rpc-Header-`

junchaowu:
Related: yarpc/yarpc#55

This also changes the TChannel inbound to always normalize incoming headers.


breerly:
This is great. @kriskowal @junchaowu possible to add Node to the mix?

abhinav:
This is ready for review. I made the HTTP headers consistent with the other languages and added some more tests on the behavior. The test is now green against all languages.

docker-compose.yml
```diff
@@ -14,10 +14,12 @@ services:
             - AXIS_CLIENT=java,node,go,python,python-sync
             - AXIS_SERVER=java,node,go,python
             - AXIS_TRANSPORT=http,tchannel
+            - AXIS_ENCODING=raw,json,thrift
 
             - BEHAVIOR_RAW=client,server,transport
             - BEHAVIOR_JSON=client,server,transport
             - BEHAVIOR_THRIFT=client,server,transport
+            - BEHAVIOR_HEADERS=client,server,transport,encoding
```
breerly: this is a heavy test :)

breerly:
Love the table tests. :shipit:

yarpc-core/src/main/java/io/yarpc/transport/RequestParser.java
```diff
@@ -55,42 +56,11 @@ public static void checkRequiredProperties(
         }
     }
 
+    @Data
```
breerly: This is so awesome.

yarpc-core/src/test/java/io/yarpc/TestUtil.java
```diff
@@ -22,14 +22,44 @@
...
+            output.put(args[i], args[i + 1]);
+        }
+        return output;
+    }
+
+    /**
+     * httpHeaders is a helper to build HttpHeaders from key-value pairs.
+     */
+    public static HttpHeaders httpHeaders(String... args) {
+        Assert.assertTrue("httpHeaders() must be called with an even number of arguments", args.length % 2 == 0);
```
breerly: that line length tho ;)

yarpc-core/src/test/java/io/yarpc/harness/behavior/headers/HeadersTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+    public String testHeader(Outbound outbound, String encoding) throws Exception {
+        String hash1 = UUID.randomUUID().toString().substring(0, 7);
+        String hash2 = UUID.randomUUID().toString().substring(0, 7);
+
+        Channel channel = Channel.builder().service("yarpc-test").caller("yarpc-test").outbound(outbound).build();
+
+        HeaderCaller caller;
+        switch (encoding) {
+            case Encoding.RAW:
```
breerly: I thought you wanted to get rid of this glossary?

yarpc-core/src/test/java/io/yarpc/TestUtil.java
abhinav: Fixed :)

yarpc-core/src/test/java/io/yarpc/harness/behavior/headers/HeadersTest.java
abhinav: I mostly translated the old tests over, but good point. Switched to not using the glossary.

yarpc-core/src/main/java/io/yarpc/transport/http/Headers.java
```diff
@@ -0,0 +1,85 @@
...
+import io.netty.handler.codec.http.HttpHeaders;
+
+import java.util.HashMap;
+import java.util.Map;
+
+class Headers {
+    /*
+     * Names of the YARPC headers
+     */
+    static final String PROCEDURE = "RPC-Procedure";
```
junchaowu: explicitly set public or private on any field.

abhinav: It's package-local. That's the behavior we want in this case since the other classes are actually accessing this.

PR #92, opened 2016-04-28
Introducing Lombok to the project

abhinav:
Note how simple classes like Channel, Config, and Context become. This is just
to give an initial taste of using this.

All stable annotations supported by Lombok are documented here:
https://projectlombok.org/features/index.html. I picked what I think is a
safer subset with unsurprising features and added it to DEVELOPING.md.

Opinions welcome.

CC @yarpc/java


DEVELOPING.md
```diff
@@ -0,0 +1,22 @@
+Getting Started
```
breerly: thanks for starting this -> will be super useful for onboarding devs!

DEVELOPING.md
```diff
@@ -0,0 +1,22 @@
...
+
+    $ make test
+
+### Lombok
+
+If you are using IntelliJ, be sure to install the [Lombok Plugin] for IntelliJ.
+When using Lombok, restrict yourself to the following annotations:
+
+    NonNull, Getter, Setter, ToString, EqualsAndHashCode, NoArgsConstructor,
+    RequiredArgsConstructor, AllArgsConstructor, Data, Value, Builder, Log*
```
breerly: Is it crazy talk to check some select IntelliJ config files? I've had limited success with this type of thing in the past.

breerly:
💯 🚢

DEVELOPING.md
abhinav: We could check in the config files but they won't provide the Lombok plugin and there's no existing way to enforce the annotations subset. It would, however, provide the project structure (although Gradle handles that pretty well already), and consistent code styling settings. I'll look into it for a different PR.

prashantv:
👍 this is sweet

abhinav:
If there are no other comments, I'm going to merge this.

PR #93, opened 2016-04-28
Check gradlew into the repo.

abhinav:
As per https://docs.gradle.org/current/userguide/gradle_wrapper.html, you can
(and should) check in `gradlew`, `gradlew.bat`, and
`gradle-wrapper.{jar,properties}` into the repo to make sure the same version
is used everywhere.

Note that Gradle generates a wrapper and jar for each subproject. Since the
jar files are all the same, I kept the one at the top level and symlinked the
others. I also deleted the `gradlewrapper` Make target because we want other
developers to always use the `gradlew` bundled with the repository. We will
run `gradle wrapper` only when we're upgrading the wrapper script with a newer
version of Gradle.


junchaowu:
lgtm

PR #94, opened 2016-04-28
Travis: Cache Thrift builds

abhinav:
This builds Thrift in `~/.thrift-build` and caches that so that we don't have to rebuild it for each commit and PR.


abhinav:
This is ready for review. Note that [this build](https://travis-ci.org/yarpc/yarpc-java/builds/126266255) used the cached, already-built Thrift rather than fetching and building it. This cut the build time down from 7 minutes to ~4 minutes.

breerly:
Hard to argue :shipit:

PR #96, opened 2016-04-28
Set up checkstyle

abhinav:
This enables checkstyle for the project and adds a very basic config. We can
iterate on it later and add more checks.

Note about build targets: `./gradlew test` runs just the tests, and `./gradlew
check` runs the tests and checkstyle (and possibly other linters/analyzers if
we set them up later).


abhinav:
CC @yarpc/java

breerly:
👍  hard to argue

junchaowu:
I was holding on adding this because I was waiting for the java support team to bring the uber checkstyle out. I wasn't sure where to find your xml but I let the patch in for a good start.

PR #99, opened 2016-05-04
Middleware (interceptors & filters)

junchaowu:
Java impl of https://github.com/yarpc/yarpc/issues/57


yarpc-example/src/main/java/io/yarpc/example/tchannel/thrift/Client.java
```diff
@@ -75,7 +75,7 @@ public static void main(String[] args) throws InterruptedException, IOException,
         } catch (Exception e) {
             e.printStackTrace();
         } finally {
-            tchannel.shutdown(false);
+            client.stop();
```
breerly: nice

yarpc-core/src/main/java/io/yarpc/Channel.java
```diff
@@ -21,18 +21,102 @@
...
+        this.service = builder.nestedService;
+        this.caller = builder.nestedCaller;
+        this.outbound = builder.nestOutbound;
+        this.interceptors = new LinkedList<>();
+        if (builder.nestInterceptors != null) {
+            this.interceptors.addAll(builder.nestInterceptors);
+        }
+    }
+
+    public String getService() {
```
breerly: no lombok?

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -34,16 +37,18 @@
 import java.util.List;
 
 public class RPC {
-    private static final Logger logger = LoggerFactory.getLogger(RPC.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(RPC.class);
```
breerly: does this make it impossible to inject a custom logger?

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -34,16 +37,18 @@
...
     private Config config;
     private String name;
     private HandlerRegistry mHandlerRegistry;
     private RPCFuture<Void> runningFuture;
+    private InterceptorManager interceptorManager;
```
breerly: Please don't name this `InterceptorManager` - what does this class actually do? Is it just a collection of interceptors or does it actually perform some sort of behavior on them?

yarpc-core/src/main/java/io/yarpc/Channel.java
junchaowu: A separate PR to change all code to use lombok will make more sense. The lombok is added in the middle of the time I was doing the middleware change

yarpc-core/src/main/java/io/yarpc/interceptor/InterceptorManager.java
```diff
@@ -0,0 +1,41 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.interceptor;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class InterceptorManager {
```
breerly: Suggest renaming to `InterceptorCollection` since this appears to be a well-defined container for multiple interceptors.

yarpc-core/src/main/java/io/yarpc/interceptor/InterceptorManager.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public InterceptorManager() {
+        this.clientInterceptors = new LinkedList<>();
+    }
+
+    public List<PhaseInterceptor> getClientInterceptors() {
+        return this.clientInterceptors;
+    }
+
+    public void addClientInterceptor(PhaseInterceptor interceptor) {
```
breerly: Can you explain the domain language used here?

What's a PhaseInterceptor vs a regular Interceptor vs a ClientInterceptor?

yarpc-core/src/main/java/io/yarpc/util/BufferUtil.java
```diff
@@ -49,4 +49,13 @@
 
         return bs;
     }
+
+    public static ByteBuffer clone(ByteBuffer original) {
```
breerly: I'm against bucketing functionality in a util namespace. The 3 methods in this class are all Buffer operations, let's move them to something more specific like `yarpc/Buffer.java` or `yarpc/ByteBuffer.java`.

yarpc-core/src/main/java/io/yarpc/Channel.java
breerly: 👍 ; i had thought we were already using it.

yarpc-core/src/main/java/io/yarpc/interceptor/ClientTranslateInterceptor.java
```diff
@@ -0,0 +1,92 @@
...
+import io.yarpc.Request;
+import io.yarpc.Response;
+import io.yarpc.encoding.Serializer;
+import io.yarpc.exception.ExceptionMessages;
+import io.yarpc.exception.InvalidInboundResponseException;
+import io.yarpc.exception.InvalidOutboundRequestException;
+
+import java.nio.ByteBuffer;
+
+public class ClientTranslateInterceptor<O> extends AbstractPhaseInterceptor {
```
breerly: docs for this class?

yarpc-core/src/main/java/io/yarpc/interceptor/InterceptorDLink.java
```diff
@@ -0,0 +1,66 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.interceptor;
+
+public class InterceptorDLink {
```
breerly: What does DLink stand for?

yarpc-core/src/main/java/io/yarpc/interceptor/ServerTranslateInterceptor.java
```diff
@@ -0,0 +1,93 @@
...
+import io.yarpc.Request;
+import io.yarpc.Response;
+import io.yarpc.encoding.Serializer;
+import io.yarpc.exception.ExceptionMessages;
+import io.yarpc.exception.InvalidInboundRequestException;
+import io.yarpc.exception.InvalidOutboundResponseException;
+
+import java.nio.ByteBuffer;
+
+public class ServerTranslateInterceptor<O> extends AbstractPhaseInterceptor {
```
breerly: Suggest renaming these to `ServerConvertingInterceptor` instead of using `Translate`.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -93,19 +99,45 @@ public void stop() {
...
+        return new Channel.Builder()
+                .service(service)
+                .caller(caller)
+                .outbound(outbound)
+                .interceptors(interceptors)
+                .build();
+    }
+
+    public void addServerInterceptor(PhaseInterceptor interceptor) throws NotImplementedException {
+        // TODO global interceptor for whole rpc instance
```
breerly: Can you create a GH issue for this. Planning on adding this next?

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -34,16 +37,18 @@
...
     private RPCFuture<Void> runningFuture;
+    private InterceptorManager interceptorManager;
 
     public RPC(Config config) {
-        this.mHandlerRegistry = new MapHandlerRegistry();
-        this.config = config;
         this.name = config.getName();
+        this.mHandlerRegistry = new MapHandlerRegistry(this.name);
+        this.config = config;
+        this.interceptorManager = new InterceptorManager();
```
breerly: think this would read better as:

```
this.interceptors = new InterceptorCollection();
```

yarpc-core/src/main/java/io/yarpc/interceptor/InterceptorManager.java
breerly: I think I understand - this class is responsible for holding onto 2 types of interceptor collections - one for server and one for client.

I think we should instead reduce this class to a `InterceptorCollection` which contains only 1 type, and then in the RPC object we would do something like:

```java
self.interceptors = new InterceptorCollection();  // server interceptors
self.filters = new InterceptorCollection();  // client interceptors
```

Then when you were acting on RPC object, we wouldn't need `setClientInterceptor` methods, you could just (in pseudo code):

```
rpc.interceptors.push(new Interceptor())  // add global server interceptors
rpc.filters.push(new Interceptor())  // add global client interceptors
```

cc @abhinav

yarpc-core/src/main/java/io/yarpc/interceptor/InterceptorDLink.java
breerly: Doubly linked list? 

Let's call this a `NodeChain` instead.

breerly:
Here are the changes I'd like to see:

## Separate terms for server and client middleware

1. All top-level interfaces for middleware should use the terminology `Middleware`. 
2. Server-side implementations of middleware should be called `Interceptors`.
3. Client-side implementations of middleware should be called `Filters`.

This allows the top-level RPC object to contain 2 collections named interceptors and filters instead of a single interceptor manager who is responsible for both client and server side mw:

```java
class RPC {
    this.interceptors = new MiddlewareCollection();
    this.filters = new MiddlewareCollection();
}
// ...
rpc.filters.push(new CachingFilter())
rpc.interceptors.push(new AuthzInterceptor())
```

This also creates room for middlware that get applied to *both* to client *and* server-side:

```java
func addMiddleware(mw Middleware)
{
    this.interceptors.push(mw)
    this.filters.push(mw)
}
```

## Use convert instead of translate

The term `translate` tends to be related i18n for most people, while the term `convert` more directly conveys what we're doing here:

> convert - to cause to change in form, character, or function.

Let's rename classes to follow this convention:

```
ConvertingInterceptor
ConvertingFilter
```

Instead of:

```
ServerTranslateInterceptor
ClientTranslateInterceptor
```

## no util package

Instead of creating a `util` package which will end up collecting all classes we didn't challenge ourselves to name, let's organize our classes based on the domain being operated on.

Instead of `yarpc/util/BufferUtil.java`, let's move them to something more specific like yarpc/Buffer.java or yarpc/ByteBuffer.java.

yarpc-core/src/main/java/io/yarpc/RPC.java
breerly: Also, as mentioned in my lengthy comment, calling server-side middleware interceptors simplifies this interaction to something like:

```
rpc.interceptors.push(new AuthzInterceptor());
```

junchaowu:
We should keep use BufferUtil in java because the Buffer.java is confusing and it will be interpreted as Buffer bean class.

breerly:
@junchaowu re:util - i could buy that, lets leave it how it is for now since it's an internal detail

yarpc-core/src/main/java/io/yarpc/Channel.java
```diff
@@ -21,18 +21,102 @@
...
-@Builder
 public class Channel {
     // TODO add per-tchannel timeout, etc.
 
     private String service;
     private String caller;
     private Outbound outbound;
+    private List<PhaseMiddleware> filters;
+
+    public Channel(Builder builder) {
```
abhinav: Why did you you remove the `@Builder` and add all this code back in? Also, the `@Getter` and `@Setter` already give you all these getters and setters.

breerly: Looks like @junchaowu reverted the lombok code here. Is your IDE setup wrong or something?

breerly:
@junchaowu this diff is pretty large - if you had to, could you split this into separate smaller PRs? Could you try and list what those PRs would be? That might help reviewing this beast :)

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -88,24 +97,78 @@ public void stop() {
...
-                inbound.stop();
-            } catch (Exception e) {
-                e.printStackTrace();
+        Map<String, Outbound> outbounds = this.config.getOutbounds();
+        if (outbounds != null) {
+            for (Map.Entry<String, Outbound> outbound : outbounds.entrySet()) {
+                try {
+                    outbound.getValue().stop();
+                } catch (Exception e) {
+                    LOGGER.debug(e.getMessage());
```
abhinav: Failures aren't `debug`-level. It's at least a `warn` if not an `error` and should include the full stack trace, not just the message.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -88,24 +97,78 @@ public void stop() {
...
             }
         }
 
+        List<Inbound> inbounds = this.config.getInbounds();
+        if (inbounds != null) {
+            for (Inbound inbound : inbounds) {
+                try {
+                    inbound.stop();
+                } catch (Exception e) {
+                    LOGGER.debug(e.getMessage());
```
abhinav: Same as above

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -88,24 +97,78 @@ public void stop() {
...
         this.runningFuture.set(null);
     }
 
     public Channel createChannel(String service) throws OutboundNotFoundException {
-        return this.createChannel(service, this.name);
+        return this.createChannel(service, this.name, this.filters);
+    }
+
+    public Channel createChannel(String service, String caller) throws OutboundNotFoundException {
+        return this.createChannel(service, caller, this.filters);
```
abhinav: Creating channels with a different caller name should not be a public API. The RPC has a name. That should be used.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -88,24 +97,78 @@ public void stop() {
...
     }
 
-    private Channel createChannel(String service, String caller) throws OutboundNotFoundException {
+    public Channel createChannel(
+            String service, String caller, List<PhaseMiddleware> interceptors) throws OutboundNotFoundException {
```
abhinav: Can channel interceptors be changed after a Channel is created? If yes, we shouldn't provide this API.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -88,24 +97,78 @@ public void stop() {
...
+        // TODO global middleware for whole rpc instance
+        // this.interceptorManager.addServerInterceptor(middleware);
+        throw new NotImplementedException("Not support yet");
+    }
+
+    public void addFilters(List<PhaseMiddleware> filters) {
+        this.filters.addAll(filters);
+    }
+
+    public void addFilter(PhaseMiddleware middleware) {
```
abhinav: These two functions can be combined into one with the use of variadic arguments. Also, you don't need to account for every possible form of calling right off the bat. Support the minimal case and we can add more later.

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -88,24 +97,78 @@ public void stop() {
...
+        this.filters.add(middleware);
+    }
+
+    public void addInterceptors(List<PhaseMiddleware> middlewares) {
+        this.interceptors.addAll(middlewares);
+    }
+
+    public void addInterceptor(PhaseMiddleware middleware) {
```
abhinav: Combine with varargs

yarpc-core/src/main/java/io/yarpc/Channel.java
junchaowu: The lombok is added after I made the file change. There is some naming difference between lombok's generated code with my builder method. In order to avoid to change every places using Channel class. I keep what I wrote.

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: Yes they can. But it doesn't affect us provide one more api to give them opportunity to set their own interceptors the first time channel got created.

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: The naming with or without "s" make more sense. So I don't use overload

yarpc-core/src/main/java/io/yarpc/RPC.java
abhinav: Actually, we should go with minimal APIs to start with. Decreasing the surface area of the initial API gives us more freedom later. Adding an API to start with means you can't remove it without making a breaking change.

yarpc-core/src/main/java/io/yarpc/RPC.java
abhinav: `addFilters(PhaseMiddleware... filters)` is equally understandable. My objection is to the two methods, not the method name.

yarpc-core/src/main/java/io/yarpc/Channel.java
abhinav: When the master branch moves ahead of your local changes, the correct resolution is to merge those changes in and resolve conflicts in your local changes, and not stomp over the changes in master. Please merge those changes in and rename your local occurrences. The only difference is that instead of,

    new Channel.Builder().foo(...).build()

You will do,

    Channel.builder().foo(...).build()

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: I buy it.

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: I disagree with this. No need to remove this api.

abhinav: Can you expand on that? What's your disagreement in particular?

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: This will give people flexibility to do things. Since in the tchannel user case, we have encountered many times people want apply their own name. We need to learn from tchannel case. At the beginning we give lots of restriction because we don't want users to customize things but later we start to make changes for them to meet each special user case. This api just give them a chance to do that.

abhinav: Except the name of the service is already known. It's the name of the RPC. The name of the RPC is the name of the caller. Starting with restricted APIs and adding what is then determined to be needed is still the way to go. `rpc.channel()` should always return a channel which has the same name as the `rpc`. That's the contract of `channel()` in all languages. After a `Channel` exists, it's completely disconnected from the `RPC` object. Any necessary customizations, including maybe per-channel-timeouts later, may be applied on that after creating it. But the primary contract of the `rpc.channel()` method needs to hold. We can maybe add a version of `channel` that allows customization, but this isn't the time to add that API. It would need more design thought than just simple type-based overloading. We might maybe want to have a `ChannelConfiguration` object. I don't know. That's not a discussion for right now. Right now we should go for the minimal working API surface area and not account for use cases that don't necessarily exist
 yet.

breerly:
@junchaowu should this PR be closed in favor of #108 ?

junchaowu:
PR is no longer valid.

PR #101, opened 2016-05-06
Add threadlocal interceptor

junchaowu:


breerly:
@yarpc/java some eyes please

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
```diff
@@ -132,6 +133,11 @@ private void parseGenericType(Handler handler) {
     public void setInterceptors(List<PhaseMiddleware> phaseMiddlewares) {
         parseMiddleware(this.handler);
         this.middlewareChain.add(phaseMiddlewares);
+        this.addBuiltInMiddle();
```
abhinav: `setInterceptors` is a public method. If it's called multiple times with the same list, you'll end up appending the `ContextInterceptor` and `InboundApplicationinterceptor` to it multiple times.

Instead of modifying user input, you can instead construct a new `middlewareChain` with the user's interceptors and your new interceptors separately for each request. You can also cache in an instance attribute if you don't want to instantiate it each time, and mark it as dirty when `setInterceptors` is called so that the next call to `handle` will update it.

junchaowu: parseMiddleware will recreate the middlewarechain. So if user call setInterceptors, it will rebuild the middleware chain inside.

yarpc-core/src/main/java/io/yarpc/context/ThreadContext.java
```diff
@@ -0,0 +1,46 @@
...
+package io.yarpc.context;
+
+import io.yarpc.Request;
+
+public class ThreadContext {
+    private static InheritableThreadLocal<ThreadContext> threadLocal = new InheritableThreadLocal<ThreadContext>() {
+        @Override
+        protected ThreadContext initialValue() {
+            return new ThreadContext();
+        }
```
abhinav: Don't you also need to override `childValue` to connect the new ThreadContext to the parent thread context's request?

yarpc-core/src/main/java/io/yarpc/middleware/ContextFilter.java
```diff
@@ -0,0 +1,42 @@
...
+import io.yarpc.handler.Handler;
+
+public class ContextFilter extends AbstractPhaseMiddleware {
+    public ContextFilter() {
+        super(Phase.APPLICATION);
+    }
+
+    @Override
+    public Response<?> handle(Request<?> request, Handler next) throws Exception {
+        Request parentRequest = ThreadContext.get().getParentRequest();
```
abhinav: Just do this as part of the AbstractInboundHandler instead of making this a middleware. Anything that's a "required interceptor" suggests that it should just be implemented in the main logic rather than in an interceptor.

yarpc-core/src/main/java/io/yarpc/middleware/ContextInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class ContextInterceptor extends AbstractPhaseMiddleware {
+    public ContextInterceptor() {
+        super(Phase.APPLICATION);
+    }
+
+    @Override
+    public Response<?> handle(Request<?> request, Handler next) throws Exception {
+        ThreadContext.get().setParentRequest(request);
+        return next.handle(request);
```
abhinav: Just do this as part of the AbstractInboundHandler instead of making this a middleware. Anything that's a "required interceptor" suggests that it should just be implemented in the main logic rather than in an interceptor.

breerly: Agreed here - that' s a good hint.

yarpc-core/src/main/java/io/yarpc/context/ThreadContext.java
junchaowu: By default it will just inherits the threadlocal from parent thread. I don't override the `childValue` since I don't need to do any special thing here now.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
```diff
@@ -54,32 +56,56 @@
...
+
+            @Override
+            public void onFailure(@Nullable Throwable t) {
+                shutdown(finalChannel, workerGroup);
+            }
+        });
         return future;
     }
 
+    private void shutdown(Channel channel, NioEventLoopGroup workerGroup) {
```
abhinav: It looks like this callback doesn't access any instance attributes, nor does the `shutdown` method. In that case there's no reason for this to be attached to the instance. Please make it a static inner class or an external class, that takes exactly the information it needs and nothing else.

```
Futures.addCallback(future, new ShutdownChannel(workerGroup, channel));

@RequiredArgsConstructor
private static final class ShutdownChannel implements ... {
    private final NioEventLoopGroup workerGroup;
    private final Chanenl channel;

    // ...
}
```

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
```diff
@@ -54,32 +56,56 @@
...
             channel.closeFuture().addListener(ChannelFutureListener.CLOSE);
         } catch (Exception e) {
             future.setException(new TransportException(e));
         }
+
+        final Channel finalChannel = channel;
+        Futures.addCallback(future, new FutureCallback<Response<ByteBuffer>>() {
+            @Override
+            public void onSuccess(@Nullable Response<ByteBuffer> result) {
+                shutdown(finalChannel, workerGroup);
```
abhinav: How is this change related to the thread local interceptor change? Do they have to be in the same PR?

yarpc-core/src/main/java/io/yarpc/handler/AbstractInboundHandler.java
abhinav: Okay. I disagree on having required interceptors and modifying user-input, but that's a separate discussion. Deleting this comment.

yarpc-core/src/main/java/io/yarpc/context/ThreadContext.java
abhinav: 👍

yarpc-core/src/main/java/io/yarpc/middleware/ContextInterceptor.java
junchaowu: I still don't buy the idea put it in main logic yet. I'd stick with the middleware approach now.

breerly: @abhinav will you file a ticket to "Remove core logic from extension points" so that we don't forget to align Java with the principles of our other implementations?

abhinav: https://github.com/yarpc/yarpc-java/issues/112

PR #108, opened 2016-05-10
yarpc middleware + error handling

junchaowu:
Please review by commits.


breerly:
@junchaowu can you resolve theses conflicts?

docker-compose.yml
```diff
@@ -4,47 +4,48 @@ services:
         image: yarpc/crossdock
         links:
             - java
-            - node
             - go
-            - python
-            - python-sync
+            - node
+#            - python
+#            - python-sync
```
breerly: python errors have made it to master - lets uncomment these before landing this PR

docker-compose.yml
```diff
@@ -4,47 +4,48 @@ services:
...
 
-            - AXIS_CLIENT=java,node,go,python,python-sync
-            - AXIS_SERVER=java,node,go,python
+            - AXIS_CLIENT=node,java,go
+            - AXIS_SERVER=node,java,go
             - AXIS_TRANSPORT=http,tchannel
             - AXIS_ENCODING=raw,json,thrift
-
+            - AXIS_ERRORS=node,go
             - BEHAVIOR_RAW=client,server,transport
```
breerly: space between behaviors and axes please

breerly:
I'd like to update the middleware interface to be the same as a handler (like Go did).

In python terms, this would look like:

```
class MiddlewareImpl:
    def __init__(self, next):
         self.next = next
    def handle(self, request):
        return next(request)
```

Instead of:

```
class MiddlewareImpl:
    def handle(self, request, next):
        return self.next(request)
```

This makes Middleware and Handlers fulfill the same contract, which will be helpful in the long run. I could see us needing to provide additional params in the future, like differentiating between `MASTER` and `SUB` requests for example. 

G

yarpc-core/src/test/java/io/yarpc/middleware/MiddlewareShortCutTest.java
```diff
@@ -0,0 +1,226 @@
...
+import java.io.IOException;
+import java.net.Inet4Address;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+public class MiddlewareShortCutTest {
+
+    private static final int httpPort = 9021;
```
breerly: suggest using `0` here

yarpc-example/src/main/java/io/yarpc/example/tchannel/json/Client.java
```diff
@@ -64,7 +64,7 @@ public static void main(String[] args) throws InterruptedException, IOException,
         } catch (Exception e) {
             e.printStackTrace();
         } finally {
-            tchannel.shutdown(false);
+            client.stop();
```
breerly: love this change :)

breerly:
🚢 

My browser can barely handle this page anymore - but it passes Crossdock tests, and that's a good starting point 👍

yarpc-core/src/main/java/io/yarpc/RPC.java
```diff
@@ -88,24 +98,59 @@ public void stop() {
...
+        return this.createChannel(service, this.name, this.filters);
     }
 
-    private Channel createChannel(String service, String caller) throws OutboundNotFoundException {
+    public Channel createChannel(String service, String caller) throws OutboundNotFoundException {
+        return this.createChannel(service, caller, this.filters);
+    }
+
+    public Channel createChannel(
+            String service, String caller, List<PhaseMiddleware> filters) throws OutboundNotFoundException {
```
abhinav: I still think these APIs should be private at the least. Customizing your caller name outside of the RPC object isn't supported yet. Lower-level APIs exist for that (i.e., build your own Channel).

yarpc-core/src/main/java/io/yarpc/middleware/Direction.java
```diff
@@ -0,0 +1,27 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.middleware;
+
+public enum Direction {
+    TtoA,
+    AtoT
```
abhinav: More descriptive names please.

yarpc-core/src/main/java/io/yarpc/middleware/MiddlewareChain.java
```diff
@@ -0,0 +1,93 @@
...
+public class MiddlewareChain {
+    private Handler endHandler;
+    private Direction direction;
+    private MiddlewareDLink interceptorDLink;
+    private MiddlewareNode phaseSeparator;
+
+    public MiddlewareChain(Direction direction, PhaseMiddleware phaseMiddleware, Handler endHandler) {
+        this.direction = direction;
+        this.interceptorDLink = new MiddlewareDLink();
+        this.phaseSeparator = new MiddlewareNode(phaseMiddleware);
```
abhinav: This `phaseSeparator` business seems unnecessary versus maintaining two lists.

breerly:
Please get @abhinav's stamp before merging

breerly:
File an issue to cover implementing rpc-wide interceptors.

yarpc-core/src/main/java/io/yarpc/middleware/MiddlewareChain.java
junchaowu: this is because if I use one list then I need my converting middler to separate `transport` and `application` layer. Of course the other way to do is two lists. In that case the first list's tail should call second list's head and it is also possible the second list doesn't exist. The situation is more complicated then just maintain a single middleware list.

yarpc-core/src/main/java/io/yarpc/RPC.java
junchaowu: I don't feel this is critical here. @breerly do you like this API or not?

junchaowu:
https://github.com/yarpc/yarpc-java/issues/109

yarpc-core/src/main/java/io/yarpc/middleware/MiddlewareChain.java
```diff
@@ -0,0 +1,93 @@
...
+        return phaseSeparator;
+    }
+
+    public void add(List<PhaseMiddleware> interceptors) {
+        for (PhaseMiddleware interceptor : interceptors) {
+            this.add(interceptor);
+        }
+    }
+
+    public <I> Response<?> handle(Request<I> request) throws Exception {
```
abhinav: If I'm reading this correctly, `MiddlewareChain` takes a series of transport and application level middlewares and adapts them into a `Handler`-like interface?

Why does this need to be aware of phase interceptor at all? Can't it use a single `List<Middleware>` and a `Handler`? Something approximately like,

```java
class MiddlewareChain {
     private List<Middleware> middlewares;
     private Handler handler;

     public Response<?> handle(Request<I> request) throws Exception {
         return new Execution().handle(request);
     }

     private class Execution implements Handler {
         private int index = 0;

         @Override
         public Response handle(Request request) throws Exception {
             if (index == middlewares.length()) {
                 return handler.handle(request);
             }
             return middlewares.get(index++).handle(request, this);
         }
     }
}
```

(You'll want the list to be an arraylist since we'll be doing random lookups in it.)

abhinav: See also the logic for `interceptorChain` in yarpc-go: https://github.com/yarpc/yarpc-go/pull/124/files#diff-60607913816ced03de51d4ebc9ad80afR67

yarpc-core/src/main/java/io/yarpc/middleware/ConvertingFilter.java
```diff
@@ -0,0 +1,93 @@
...
+import io.yarpc.exception.ExceptionMessages;
+import io.yarpc.exception.InvalidInboundResponseException;
+import io.yarpc.exception.InvalidOutboundRequestException;
+import io.yarpc.handler.Handler;
+
+import java.nio.ByteBuffer;
+
+public class ConvertingFilter<O> extends AbstractPhaseMiddleware {
+    private final Class<O> responseBodyType;
+    private Serializer serializer;
```
abhinav: Use the full type of the serializer here if possible.

```java
class ConvertingFilter<B> extends ... {
    private final Class<? extends B> responseBodyType;
    private Serializer<B> serializer;
}
```

yarpc-core/src/main/java/io/yarpc/middleware/MiddlewareChain.java
junchaowu: Because I am totally replying on the chain, not the list here. This is the difference implementation for the same thing.

abhinav: Yes, it's ultimately similar, but one is significantly more complex than the other. Complexity increases the likelihood of introducing bugs when the code is changed later. If one solution lets you avoid implementing custom data structures, that's preferable.

yarpc-core/src/main/java/io/yarpc/RPC.java
breerly: The other langs don't allow passing caller when creating a channel - lets keep it consistent.

yarpc-core/src/main/java/io/yarpc/middleware/ConvertingFilter.java
junchaowu: https://github.com/yarpc/yarpc-java/issues/110 create issue for this.  I believe this kind of thing appear in other places as well.

yarpc-core/src/main/java/io/yarpc/middleware/Direction.java
junchaowu: I will walk through whole project to add java docs. https://github.com/yarpc/yarpc-java/issues/111

yarpc-core/src/main/java/io/yarpc/middleware/ConvertingFilter.java
abhinav: I think when you're adding new code, try to follow what has already been established as something we want to do. Deferring to a time when we're going to go through all the code and fix it is not the best strategy, especially for all this new code. In this specific case, if you can already use proper types, use them. If you can't, *then* we fix it later.

yarpc-core/src/main/java/io/yarpc/middleware/Direction.java
abhinav: This isn't about docs. This is about a descriptive name for the two enums. What about `TransportToApplication` and `ApplicationToTransport`?

breerly: +1 to `TransportToApplication` and `ApplicationToTransport` - lets prefer domain-terms over abbreviations.

abhinav:
Is there a way we avoid moving core library logic into middleware? Given that middleware is basically untyped (`Response<?> handle(Request<?> request, Handler next)`), we're losing a lot of type safety in core library logic around what the requests and responses contain. I dislike that we have to cast during critical points like "convert to/from application request/response". Previously these things were done in fairly safe parts of the code where the exact types of request and response bodies were known at compile time. Moving this into the middleware means that we can't fix this even with #110 since middlewares will (presumably) have to remain untyped.

breerly:
@abhinav what are the dangers of losing type-safety in these core parts? What could go wrong, rather.

I agree that we probably shouldn't be implementing our own library in-terms of extension-points - those should be reserved for client code only. This more than anything makes me want to pull this functionality out.

This will probably be a decent refactor of this PR, potentially pushing it out days/weeks - can you make a case for why we should be keeping type safety around these requests and avoiding casts?

abhinav:
I mean we don't *have* to have type safety around core library logic but given that this is intended to be a heavily used library, we *should* try to follow practices that decrease the likelihood of runtime bugs.

We have `ConvertingFilter` and `ConvertingInterceptor`, for example, that convert between the application/transport requests (including calling the serializer). 

```java
    @Override
    public Response<?> handle(Request<?> request, Handler next) throws Exception {
        Request<ByteBuffer> transportRequest = applicationToTransport(request);
        Response<ByteBuffer> transportResponse = (Response<ByteBuffer>) next.handle(transportRequest);
        return transportToApplication(transportResponse, transportRequest);
    }
```

Because this is a middleware, you have no guarantee on what the `Request` contains and what types the next `Handler` expects. Any mismatches will result in runtime failures. It will be very easy for someone to come in and mess up the types there. Part of this conversion was done in `RPCClient` previously, inside a scope where the types of the request and response bodies were known exactly.

```java
    protected <I extends B, O extends B> RPCFuture<Response<O>> send(
            Request<I> request,
            Class<O> responseBodyType
    ) throws Exception {
        // transfer application level request to transport level request
        Request<ByteBuffer> transportRequest = Request.toByteBufferRequest(request, getSerializer());
```

I don't know what form the changes I'm asking for will take, but @junchaowu please look into moving core logic out of the middleware rather than into it.

junchaowu:
I don't see any reason this is bad. The network RPC call flow is a pipeline at the end. The request is passed from one middleware to another. As long as our yarpc can make sure the sequence of the middleware is right. Event though the request is unsafe type, the error handling can deal with it. Even though I put the logic like convertingFilter into rpcclient, I still don't know if the request object I get through the interceptor is right or not. And it will make the code complicated. All in all the error handling will capture the errors and middleware chain will make us easy to add new feature into the yarpc.

junchaowu:
Add to my point, middlewares added by users are part of users' code. If they didn't write correctly, as long as our yarpc can handle the exception and propagate right error to the user side, let them know which part is wrong, I am okay with it. Follow this principle, I don't see program with this.

breerly:
@junchaowu - talked with @abhinav and @kriskowal -

We're good to 🚢  this PR now. When you're in tomorrow, we wan't to talk about moving core functionality out of extension points, as this is a stance we want to take with all of the YARPC libs.

PR #113, opened 2016-05-12
refactor the http transport to close connection after each request

junchaowu:
@yarpc/java


blampe:
Don't we want persistent connections?

junchaowu:
This is http transport. We are using http1.1

blampe:
Got it thanks.

yarpc-example/src/main/java/io/yarpc/example/middleware/cache/client/transport/Client.java
```diff
@@ -48,11 +47,7 @@ public static void main(String[] args) throws InterruptedException, OutboundNotF
         RPC rpc = new RPC(new Config(
                 "testServer",
                 new HashMap<String, Outbound>() {{
-                    try {
-                        this.put("testServer", new HTTPOutbound("127.0.0.1", 9000));
-                    } catch (IOException e) {
```
breerly: love that we dont need to catch this now :)

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
```diff
@@ -115,4 +121,34 @@ private Bootstrap bootstrap(
         bootstrap.handler(httpClientInitializer);
         return bootstrap;
     }
+
+    private class ResponseFutureCallback implements FutureCallback<Response<ByteBuffer>> {
```
abhinav: static?

breerly:
lgtm - the tests not having to catch the ioexception is better.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
abhinav: Also, can you give it a more descriptive name? Rather than where it's used, what it does. Like, `ShutdownChannelCallback`? We don't even necessarily need to add `Callback` to the name.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
```diff
@@ -54,32 +56,36 @@
...
+            Bootstrap bootstrap = this.bootstrap(workerGroup, new HTTPClientInitializer());
+            channel = bootstrap.connect(this.host).sync().channel();
             channel.pipeline().addLast("custom", new HTTPClientInboundHandler(future));
             this.sendRequest(channel, transportRequest);
             channel.closeFuture().addListener(ChannelFutureListener.CLOSE);
         } catch (Exception e) {
             future.setException(new TransportException(e));
         }
+
+        final Channel finalChannel = channel;
```
abhinav: You don't need to do this line or the `final` above anymore

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
```diff
@@ -115,4 +121,34 @@ private Bootstrap bootstrap(
         bootstrap.handler(httpClientInitializer);
         return bootstrap;
     }
+
+    private class ResponseFutureCallback implements FutureCallback<Response<ByteBuffer>> {
+        private final Channel channel;
+        private final NioEventLoopGroup workerGroup;
+
+        public ResponseFutureCallback(Channel channel, NioEventLoopGroup workerGroup) {
```
abhinav: Can you mark Channel `@Nullable` here?

abhinav:
I think @blampe is right. We'll lose persistent connections. It sounds like Netty's default HTTP server/client supports persistent HTTP connections by default.

yarpc-core/src/main/java/io/yarpc/transport/http/HTTPOutbound.java
```diff
@@ -54,32 +56,35 @@
...
-    public HTTPOutbound(String ip, int port) throws IOException, InterruptedException {
+
+    public HTTPOutbound(String ip, int port) {
         this.host = new InetSocketAddress(ip, port);
-        this.bootstrap = this.bootstrap(this.workerGroup, new HTTPClientInitializer());
     }
 
     @Override
     public RPCFuture<Response<ByteBuffer>> send(Request<ByteBuffer> transportRequest) throws TransportException {
+        final NioEventLoopGroup workerGroup = new NioEventLoopGroup();
```
abhinav: You don't need this final eitehr

abhinav:
Missed a `final`. LGTM besides that.

PR #115, opened 2016-05-13
Add threadlocal interceptor

junchaowu:
The threadlocal strategy  is used here for future context propagation feature. Threadlocal acts like "request context" in python , it can save the current request on current thread or child thread. In my test, I have written TestContextFilter, which get the parent request from threadlocal and modified the caller name for child request transparently. In the future, this can be adapted for tracing context propagation.


breerly:
@junchaowu care to leave a description of what this is and why we need it?

junchaowu:
updated the description.

breerly:
lgtm

PR #121, opened 2016-07-26
Require encoding over the wire

abhinav:
This changes yarpc-go to require the encoding over the wire. Incoming requests
are considered bad requests if they don't have an encoding or if the encoding
does not match the encoding on the handler.

Related: yarpc/yarpc-go#279 yarpc/yarpc-node#77 yarpc/yarpc-python#88


prashantv:
lgtm

PR #127, opened 2017-01-17
travis: Update to match yarpc-go

abhinav:
This splits the CI job into a test and crossdock target, running concurrently.
The test target tests against both, JDK 7 and 8. We need sudo for crossdock
only.

We push the crossdock image only if the DOCKER_EMAIL, DOCKER_USER and
DOCKER_PASS environment variables are set. (They are unset for pull requests
made from forks of the repo.)


abhinav:
We need https://github.com/yarpc/yarpc-node/pull/83 before this will pass

PR #132, opened 2017-01-18
Add yarpc-{encoding-raw,integration-tests} modules

abhinav:
This pulls the YARPC Raw encoding into its own package and it moves tests in
yarpc-core that depend on the encoding into the yarpc-integration-tests
package.

@bpon


bpon:
:shipit:

PR #138, opened 2017-01-18
tchannel: Drop transport

abhinav:
This deletes the TChannel transport from YARPC. To make this build, I also
ended up moving the Thrift test and codegen to the yarpc-encoding-thrift
package.


bpon:
:shipit:

PR #139, opened 2017-01-18
Include test coverage for subprojects

bpon:
cc @abhinav


build.gradle
```diff
@@ -111,3 +111,35 @@ subprojects {
...
+            enabled = true
+            destination = "${buildDir}/reports/jacoco/jacocoRootReport/jacocoRootReport.xml"
+        }
+    }
+}
+
+jacocoRootReport {
+    afterEvaluate {
+        classDirectories = files(classDirectories.files.collect {
+            fileTree(dir: it, exclude: ['**/generated/**', '**/*Test*', '**/exception/**'])
```
bpon: It sucks having to do this at the root project and all subprojects, but I can't find a better way right now.

PR #146, opened 2017-01-20
core: Add middleware to dispatcher

bpon:
Also move middleware to its own package.

@abhinav


yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
```diff
@@ -112,6 +126,26 @@ public DispatcherBuilder outbound(String serviceName, OnewayOutbound outbound) {
             return this;
         }
 
+        public DispatcherBuilder inboundMiddleware(UnaryInboundMiddleware unaryInboundMiddleware) {
```
abhinav: `@NonNull`? Also, should the Builder use a list instead so that it can be chained into one at `build` time?

yarpc-core/src/main/java/io/yarpc/core/transport/middleware/UnaryInboundMiddleware.java
```diff
@@ -19,8 +19,15 @@
...
+import io.yarpc.core.transport.TransportException;
+import io.yarpc.core.transport.TransportRequest;
+import io.yarpc.core.transport.TransportResponseWriter;
+import io.yarpc.core.transport.UnaryHandler;
 
 public interface UnaryInboundMiddleware {
-    void handle(TransportRequest req, TransportResponseWriter resWriter, UnaryOutbound next) throws TransportException;
+    UnaryInboundMiddleware NOP = (req, resWriter, next) -> next.handle(req, resWriter);
+
+    void handle(TransportRequest req, TransportResponseWriter resWriter, UnaryHandler next) throws TransportException;
```
abhinav: Why a separate package?

yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
```diff
@@ -122,7 +164,31 @@ public Dispatcher build() {
                             .build()
                     )
                     .collect(Collectors.toMap(ClientConfig::getService, Function.identity()));
-            return new Dispatcher(name, inbounds, outbounds);
+            UnaryInboundMiddleware unaryInboundMiddlewareChain = unaryInboundMiddlewares.stream()
+                    .reduce((chain, middleware) ->
+                            (r, w, n) -> chain.handle(r, w, new UnaryHandlerWithMiddleware(n, middleware))
+                    )
```
abhinav: We can do this later: Instead of reducing, might be cleaner to use explicit
`Chain*Middleware` classes. Something like,

```java
class ChainUnaryOutboundMiddleware implements UnaryOutboundMiddleware {
  private final List<UnaryOutboundMiddleware> items;

  public TransportResponse call(TransportRequest request, UnaryOutbound out) {
    return new ChainOutbound(out).call(request)
  }
  
  @RequiredArgsConstructor
  private class ChainOutbound implements Outbound {
    private final Outbound outbound;
    private int index = 0;
    
    public TransportResponse call(TransportRequest request) {
      if (index == items.size()) {
        return outbound.call(request)
      }
      OutboundMiddleware next = items.get(this.index++);
      return next.call(request, this);
    }
  }
}
```

yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
```diff
@@ -160,6 +226,24 @@ public static OutboundConfig oneway(OnewayOutbound onewayOutbound) {
         private OnewayOutbound onewayOutbound;
     }
 
+    @Data
```
abhinav: `@Data` generates toString, equals, and hashCode as well. Those don't apply here. You'll have to use `@Getter`, `@Setter`, `@RequiredArgsConstructor` explicitly.

PR #148, opened 2017-01-24
transport: Make TransportRequest an interface

abhinav:
This makes TransportRequest an interface and adds a DefaultTransportRequest
implementation. The implementation provides a Builder which validates that the
request is valid.


yarpc-core/src/main/java/io/yarpc/core/transport/TransportRequest.java
```diff
@@ -21,35 +21,97 @@
...
-    private String routingDelegate;
+    /**
+     * Contents of the request.
+     */
+    InputStream getBody();
 
-    @NonNull
-    private InputStream body;
+    /**
+     * Amount of time within which the request is expected to finish.
```
bpon: For all the optional fields, it's probably worth explicitly mentioning what a null value means, e.g. "..., or null for no timeout".

abhinav: clarified null behavior for optionals. all others are required to be non-null.

PR #162, opened 2017-01-25
Delete unused exception types and ExceptionMessages

abhinav:


yarpc-core/src/main/java/io/yarpc/core/transport/TransportException.java
```diff
@@ -21,9 +21,7 @@
  */
 package io.yarpc.core.transport;
 
-import io.yarpc.core.YARPCException;
-
-public class TransportException extends YARPCException {
+public class TransportException extends Exception {
```
breerly: Shouldn't the top of the tree be a YARPC exception instead of a plain Exception?

abhinav: We'll be adding back YARPC-top-level exception types when we get to error handling/hierarchy. For now, transport errors will inherit TransportException which can then inherit YARPCException/whatever.

PR #172, opened 2017-01-25
http/inbound: Don't use `available()` to copy bytes

abhinav:
`available()` is unreliable for this purpose. We should use `read()` until it
returns `-1` when trying to read an `InputStream` to exhaustion.


.travis.yml
```diff
@@ -76,6 +63,13 @@ script:
       ./gradlew coveralls
       ;;
     crossdock)
+      docker version
+      sudo rm -f /usr/local/bin/docker-compose
+      curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+      chmod +x docker-compose
+      sudo mv docker-compose /usr/local/bin
+      docker-compose version
```
breerly: Looks like this should be in another PR?

abhinav: Yeah, I was going to make that into a separate PR once I verified it solved the problem. Looks like it didn't.

PR #182, opened 2017-01-26
json: Revive yarpc-encoding-json project

bpon:
Needs rebasing after #180


yarpc-core/src/main/java/io/yarpc/core/encoding/OnewayDataHandler.java
```diff
@@ -19,22 +19,21 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package io.yarpc.encoding.json;
+package io.yarpc.core.encoding;
```
bpon: LOL I think git detects this as a move rather than a new file because of the licensing text.

settings.gradle
```diff
@@ -1,7 +1,7 @@
 rootProject.name = 'yarpc'
 include 'yarpc-core'
 // include 'yarpc-crossdock' TODO: re-enable
-// include 'yarpc-encoding-json' TODO: re-enable
+include 'yarpc-encoding-json'
```
abhinav: top level build.gradle needs a similar change

yarpc-core/src/main/java/io/yarpc/core/encoding/OnewayDataHandler.java
```diff
@@ -19,22 +19,21 @@
...
-
-  public void setA(String a) {
-    this.a = a;
-  }
+/**
+ * A data handler for a oneway request.
+ *
+ * @param <T> The request data type
+ */
+public interface OnewayDataHandler<T> {
```
abhinav: Data? How about Request? OnewayRequestHandler?

yarpc-core/src/main/java/io/yarpc/core/encoding/UnaryDataHandler.java
```diff
@@ -19,16 +19,23 @@
...
-  public JsonInboundHandler(Handler handler) {
-    super(new JsonSerializer(), handler);
-  }
+/**
+ * A data handler for a unary request.
+ *
+ * @param <T> The request data type
+ * @param <R> The response data type
+ */
+public interface UnaryDataHandler<T, R> {
```
abhinav: Same as above

abhinav: Nit: For this case, I would be tempted to use the I-O distinction that Java type parameters have for things that are input/output. So, `UnaryRequestHandler<I, O>` to indicate that I is the input type (request) and O is the output type (response).

PR #185, opened 2017-01-26
json: Implement JSON client

bpon:
@abhinav @vivekrsharma 

Tests coming, but just putting this out there first.


yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -0,0 +1,107 @@
...
+   * @param procedure The procedure name
+   * @param reqBody The serializable request body
+   * @param resClass The response type
+   * @param <R> The response type
+   * @return The response data
+   * @throws JsonEncodingException If an error occurred while encoding the request or decoding the
+   *     response
+   * @throws TransportException If an error occurred during transport
+   */
+  public <R> R callUnary(Context ctx, String procedure, Object reqBody, Class<R> resClass)
```
abhinav: With Java 8 type inference, you don't need to take the response class as an argument. I don't know if there's a way to go from that to the `Class` object (for `readValue`), but it's worth looking into.

bpon: That's the problem.  AFAIK You can't get a `Class` object from a generic type, and Jackson needs that.

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -0,0 +1,107 @@
...
+   * @param ctx The request context
+   * @param procedure The procedure name
+   * @param reqBody The serializable request body
+   * @param <R> The ack type
+   * @return An ack
+   * @throws JsonEncodingException If an error occurred while encoding the request or decoding the
+   *     response
+   * @throws TransportException If an error occurred during transport
+   */
+  public <R extends Ack> R callOneway(Context ctx, String procedure, Object reqBody)
```
abhinav: Why is Ack a generic? It should just be 'Ack'.

PR #192, opened 2017-01-28
thrift: Implement low-level thrift encoding api

bpon:
This is not the user-facing API.  Still need to generate client classes from IDL.

@abhinav @vivekrsharma


yarpc-encoding-thrift/src/main/java/io/yarpc/encoding/thrift/OnewayThriftTransportHandler.java
```diff
@@ -21,30 +21,35 @@
...
+@RequiredArgsConstructor
+class OnewayThriftTransportHandler<T extends TBase> implements OnewayHandler {
+  @NonNull private final OnewayRequestHandler<T> handler;
+  @NonNull private final Class<T> klass;
 
   @Override
-  public String getEncoding() {
-    return Encoding.THRIFT;
+  public void handleOneway(TransportRequest req) throws TransportException {
+    TDeserializer deserializer = new TDeserializer(new TCompactProtocol.Factory());
```
abhinav: All our systems use TBinaryProtocol.

abhinav: Also, this should probably be injected into this class from the class that
instantiates it.

yarpc-encoding-thrift/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
```diff
@@ -21,31 +21,95 @@
...
+   * @throws ThriftEncodingException If an error occurred while encoding the request or decoding the
+   *     response
+   */
+  public <R extends TBase> R callUnary(
+      Context ctx, String procedure, TBase reqBody, Class<R> resClass)
+      throws ThriftEncodingException {
+    TransportRequest req = buildTransportRequest(ctx, procedure, reqBody);
+    try {
+      TransportResponse res = clientConfig.getUnaryOutbound().callUnary(req);
+      TDeserializer deserializer = new TDeserializer(new TCompactProtocol.Factory());
```
abhinav: Binary. This one wouldn't be injected because we don't yet support using
anything else. Passing anything but Binary is going to break cross-language
compatibility.

PR #193, opened 2017-01-28
raw: add javadocs for raw encoding

vivekrsharma:
Add javadocs for public methods.


yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -42,6 +42,14 @@
 
   private static final String ENCODING = "raw";
 
+  /**
+   * Make a unary procedure call.
```
abhinav: Use present tense verbs for javadocs:

    Makes a unary procedure call.

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -42,6 +42,14 @@
...
+   *
+   * @param ctx The request context
+   * @param procedure The procedure name
```
abhinav: Start with lower case after the `@`-tags. Also, drop redundant 'the's:

    @param ctx request context

`@`-tags don't need to flow like sentences because they're just itemized at
the bottom

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -57,6 +65,14 @@
     }
   }
 
+  /**
+   * Makes a oneway procedure call.
+   *
+   * @param ctx The request context
+   * @param procedure The procedure name
+   * @param reqBody The serializable request body
+   * @return An ack
```
abhinav: Don't need to add a `@return` clause if it's self explanatory or clarified in the main docblock

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/OnewayRawTransportHandler.java
```diff
@@ -36,6 +36,12 @@
 
   @NonNull private final OnewayRequestHandler<byte[]> inbound;
 
+  /**
+   * Makes a oneway procedure call.
+   *
+   * @param req The request object
+   * @throws TransportException If an error occurred while sending the request
```
abhinav: Unnecessary `@throws` clause

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/OnewayRawTransportHandler.java
```diff
@@ -36,6 +36,12 @@
 
   @NonNull private final OnewayRequestHandler<byte[]> inbound;
 
+  /**
+   * Makes a oneway procedure call.
+   *
+   * @param req The request object
```
abhinav: @param req transport request to handle

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/OnewayRawTransportHandler.java
```diff
@@ -36,6 +36,12 @@
 
   @NonNull private final OnewayRequestHandler<byte[]> inbound;
 
+  /**
+   * Makes a oneway procedure call.
```
abhinav: This doesn't make a oneway procedure call, it handles one.

```
Handles an incoming oneway procedure call. The request is forwarded to the
underlying request handler.
```

Also, this class doesn't need too much documentation since none of it is
public. This is a package-private class that just adapts the user-defined
OnewayRequestHandler to a OnewayTransportHandler.

We care more about documentation on public APIs than private.

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
abhinav: Also, this class is part of the public API. Document all public methods and the class itself.

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -57,6 +65,14 @@
     }
   }
 
+  /**
+   * Makes a oneway procedure call.
```
abhinav: Try something like,

```
Sends a raw oneway YARPC request over this client.

@param procedure name of the procedure to call
@param reqBody contents of the request

@return an Ack if the request was accepted by the transport
```

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/UnaryRawTransportHandler.java
```diff
@@ -37,6 +37,13 @@
 
   @NonNull private UnaryRequestHandler<byte[], byte[]> inbound;
 
+  /**
+   * Makes a unary procedure call.
```
abhinav: This does not make a call, it handles it.

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
vivekrsharma: Done

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
vivekrsharma: Done

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
vivekrsharma: Done

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/OnewayRawTransportHandler.java
vivekrsharma: Fixed. will add documentation only for public APIs going forward.

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/UnaryRawTransportHandler.java
vivekrsharma: Done

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/OnewayRawTransportHandler.java
vivekrsharma: Fixed

PR #195, opened 2017-01-30
raw: Use Reactive APIs

abhinav:
This changes the Raw encoding implementation to use the reactive APIs defined
in #194.

Depends on #194.


yarpc-encoding-raw/src/test/java/io/yarpc/encoding/raw/TestUtils.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Converts a series of strings into a stream of byte buffers. Each string will be a separate byte
+   * buffer in the returned stream.
+   *
+   * @param chunks zero or more strings to emit in the stream
+   * @return stream of byte buffers
+   */
+  static Observable<ByteBuffer> byteStream(String... chunks) {
+    return Observable.fromArray(chunks).map(s -> ByteBuffer.wrap(s.getBytes()));
```
bpon: UTF-8?

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/UnaryRawTRansportHandler.java
```diff
@@ -21,32 +21,35 @@
...
-import java.io.IOException;
+import java.nio.ByteBuffer;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 
 @RequiredArgsConstructor
-class UnaryRawTransportTransportHandler implements UnaryTransportHandler {
-
-  @NonNull private UnaryRequestHandler<byte[], byte[]> inbound;
+class UnaryRawTRansportHandler implements UnaryTransportHandler {
```
bpon: Looks like the file was accidentally renamed.

abhinav: The rename was intentional but the "TR" was not. UnaryRawTransportTransportHandler just reads wrong.

PR #196, opened 2017-01-31
http: Use Reactive APIs

abhinav:
This changes the HTTP transport implementation to conform to the new reactive
APIs defined in #194. This allows us to make better use of Netty's
asynchronous capabilities.

Note that currently the implementation uses chunked encoding exclusively. We
need to optimze the case where the length is known in advance because that is
going to be the majority use case.

Depends on #194.


yarpc-transport-http/src/main/java/io/yarpc/transport/http/Channels.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+      observer.onError(cause);
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+      if (klass.isInstance(msg)) {
+        @SuppressWarnings("unchecked")
+        T message = (T) msg;
```
bpon: I recently discovered `Class.cast(Object)`, which would remove the need for `SuppressWarnings`.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -61,7 +70,7 @@
 
   private NioEventLoopGroup bossGroup;
   private NioEventLoopGroup workerGroup;
-  private Channel channel;
+  @Getter private Channel channel;
```
bpon: When is `getChannel` called?

abhinav: Derp. Leftover from testing.

PR #207, opened 2017-02-03
thrift/json: Use Reactive APIs

bpon:


yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -57,15 +58,16 @@
...
-    try {
-      return objectMapper.readValue(res.getBody(), resClass);
-    } catch (IOException e) {
-      throw new JsonEncodingException(e);
-    }
+    return clientConfig
+        .getUnaryOutbound()
+        .callUnary(req)
+        .toObservable()
+        .flatMap(TransportResponse::getBody)
```
abhinav: Can drop the `toObservable`:

    callUnary(req).flatMapObservable(TransportResponse::getBody)

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/JsonClient.java
```diff
@@ -57,15 +58,16 @@
...
+        .map(resBody -> objectMapper.readValue(resBody.array(), resClass))
```
abhinav: You're just getting chunks of the body right now. You need to accumulate them.
The Raw encoding created a helper to accumulate a `Observable<ByteBuffer>`
into a `byte[]`.

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/OnewayJsonTransportHandler.java
```diff
@@ -22,30 +22,31 @@
...
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
+  public Completable handleOneway(TransportRequest req) throws TransportException {
+    Context ctx = Context.fromRequest(req);
+    return req.getBody()
+        .map(
+            reqBody -> {
+              T reqData = objectMapper.readValue(reqBody.array(), klass);
+              return handler.handleOneway(ctx, reqData);
```
abhinav: Same as above

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/UnaryJsonTransportHandler.java
```diff
@@ -22,33 +22,47 @@
...
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
+  public Single<TransportResponse> handleUnary(TransportRequest req) throws TransportException {
+    Context ctx = Context.fromRequest(req);
+    return req.getBody()
+        .map(
+            reqBody -> {
+              I reqData = objectMapper.readValue(reqBody.array(), klass);
+              return handler.handleUnary(ctx, reqData);
```
abhinav: This will call the handler for each chunk of bytes in the buffer rather than accumulating them and calling the handler once.

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/UnaryJsonTransportHandler.java
```diff
@@ -22,33 +22,47 @@
...
+            })
+        .flatMapSingle(Single::fromFuture)
+        .map(
+            resData -> {
+              TransportResponse res =
+                  DefaultTransportResponse.builder()
+                      .headers(new Headers(ctx.getHeaders())) // TODO Fix headers
```
abhinav: ctx contains the request headers. we'll expose a way for users to write response headers to the context which we will access here

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/UnaryJsonTransportHandler.java
```diff
@@ -22,33 +22,46 @@
...
-    }
+  public Single<TransportResponse> handleUnary(TransportRequest req) throws TransportException {
+    Context ctx = Context.fromRequest(req);
+    return ObservableByteStream.accumulate(req.getBody())
+        .map(
+            reqBody -> {
+              I reqData = objectMapper.readValue(reqBody, klass);
+              return handler.handleUnary(ctx, reqData);
+            })
+        .flatMap(Single::fromFuture)
```
abhinav: Nit: Instead of `Single<Future<?>>`, then `Single<?>`, consider folding the map and flatMap into one:

```
ObservableByteStream.accumulate(req.getBody())
    .flatMap(reqBody -> {
        ...
        return Single.fromFuture(handle.handleUnary(ctx, reqData));
    })
    .map(resData ->
        DefaultTransportResponse.builder
            .body(...)
            .build());
```

yarpc-encoding-thrift/src/main/java/io/yarpc/encoding/thrift/UnaryThriftTransportHandler.java
```diff
@@ -50,18 +52,26 @@ public UnaryThriftTransportHandler(UnaryRequestHandler<I, O> handler, Class<I> k
...
+    Context ctx = Context.fromRequest(req);
+    return ObservableByteStream.accumulate(req.getBody())
+        .map(
+            resBody -> {
+              I reqData = klass.newInstance();
+              TDeserializer deserializer = new TDeserializer(protocolFactory);
+              deserializer.deserialize(reqData, resBody);
+              return handler.handleUnary(ctx, reqData);
+            })
+        .flatMap(Single::fromFuture)
```
abhinav: same nit as above

PR #208, opened 2017-02-07
Use CompletionStage instead of Future

bpon:
All clients and handlers now return CompletionStage instances instead of Futures or ListenableFutures.


yarpc-core/src/test/java/io/yarpc/core/encoding/ObservablesTest.java
```diff
@@ -0,0 +1,83 @@
...
+    single.subscribe(
+        new DisposableSingleObserver<String>() {
+          @Override
+          public void onSuccess(String obj) {
+            fail();
+          }
+
+          @Override
+          public void onError(Throwable ex) {
+            fail();
```
abhinav: Please include a message for bare failures like this

yarpc-core/src/test/java/io/yarpc/core/encoding/ObservablesTest.java
```diff
@@ -0,0 +1,83 @@
...
+        });
+    CompletableFuture<String> future = Observables.toCompletionStage(single).toCompletableFuture();
+
+    try {
+      future.get(1, TimeUnit.SECONDS);
+      fail();
+    } catch (ExecutionException e) {
+      if (!(e.getCause() instanceof TestException)) {
+        fail();
+      }
```
abhinav: assertTrue?

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/OnewayJsonTransportHandler.java
```diff
@@ -43,11 +43,11 @@
...
     return ObservableByteStream.accumulate(req.getBody())
-        .map(
+        .flatMapCompletable(
             reqBody -> {
               T reqData = objectMapper.readValue(reqBody, klass);
-              return handler.handleOneway(ctx, reqData);
-            })
-        .flatMapCompletable(Completable::fromFuture);
+              return Completable.fromFuture(
+                  handler.handleOneway(ctx, reqData).toCompletableFuture());
```
abhinav: This relies on `future.get()`. I think we want to implement a callback-based
Completable instead.

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/UnaryJsonTransportHandler.java
```diff
@@ -50,7 +50,7 @@
         .flatMap(
             reqBody -> {
               I reqData = objectMapper.readValue(reqBody, klass);
-              return Single.fromFuture(handler.handleUnary(ctx, reqData));
+              return Single.fromFuture(handler.handleUnary(ctx, reqData).toCompletableFuture());
```
abhinav: Same as above

yarpc-core/src/main/java/io/yarpc/core/encoding/Observables.java
```diff
@@ -0,0 +1,92 @@
...
+package io.yarpc.core.encoding;
+
+import io.reactivex.Completable;
+import io.reactivex.Single;
+import io.reactivex.observers.DisposableSingleObserver;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import lombok.Setter;
+
+public final class Observables {
```
abhinav: nit: all public methods in this class relate more to CompletionStages than Observables. Maybe it (and the methods) should be renamed?

```
class CompletionStages {
    <T> CompletionStage<T> fromSingle(Single<T> single);
    <T> Single<T> toSingle(CompletionStage<T> stage);
    Completable toCompletable(CompletionStage<?> stage);
}
```

yarpc-core/src/main/java/io/yarpc/core/encoding/Observables.java
```diff
@@ -0,0 +1,92 @@
...
+    future.setSingleObserver(singleObserver);
+    return future;
+  }
+
+  /** Converts a {@link CompletionStage} to a {@link Single}. */
+  public static <T> Single<T> singleFromCompletionStage(CompletionStage<T> stage) {
+    return Single.create(
+        emitter ->
+            stage.whenComplete(
+                (result, ex) -> {
```
abhinav: Nit: Consider moving the second closure into a nested class.

    Single.create(emitter -> stage.whenComplete(new SingleCompletionHandler(emitter)));

yarpc-core/src/main/java/io/yarpc/core/encoding/Observables.java
```diff
@@ -0,0 +1,92 @@
...
+                  }
+                }));
+  }
+
+  /** Converts a {@link CompletionStage} to a {@link Completable}. */
+  public static Completable completableFromCompletionStage(CompletionStage<?> stage) {
+    return Completable.create(
+        emitter ->
+            stage.whenComplete(
+                (result, ex) -> {
```
abhinav: Same as above:

    Completable.create(
      emitter -> stage.whenComplete(new CompletableCompletionHandler(emitter))
    );

yarpc-core/src/main/java/io/yarpc/core/encoding/Observables.java
```diff
@@ -0,0 +1,92 @@
...
+                  if (ex != null) {
+                    emitter.onError(ex);
+                  } else {
+                    emitter.onComplete();
+                  }
+                }));
+  }
+
+  private static class SingleCompletableFuture<T> extends CompletableFuture<T> {
+    @Setter private DisposableSingleObserver<T> singleObserver;
```
abhinav: This doesn't need to be a single observer. A `Disposable` will suffice. And then this class can be renamed to `DisposingCompletableFuture` or similar.

PR #209, opened 2017-02-07
Exclude thrift-generated classes from code coverage

bpon:
Previously, we were excluding thrift-generated classes only from the yarpc-core project.  They need to be excluded from the top-level report.


build.gradle
```diff
@@ -196,6 +196,11 @@ task jacocoRootReport(type: JacocoReport) {
             destination = "${buildDir}/reports/jacoco/jacocoRootReport/jacocoRootReport.xml"
         }
     }
+    doFirst {
+        classDirectories = files(classDirectories.files.collect {
+            fileTree(dir: it, exclude: ['**/generated/*'])
```
abhinav: I think generated code goes under `generated-sources/`, not `generated/`.

bpon: It's the package name, not the directory name.

PR #225, opened 2017-02-14
http: Support non-chunked transfer encoding

### LINKED ISSUE #197: [rx] http: Support non-chunked encoding
With https://github.com/yarpc/yarpc-java/pull/196, the HTTP server and client use chunked encoding exclusively. That's undesirable for the majority use case where the content length is known in advance.

To make this better, we may need to replace `Observable<ByteBuffer>` with a different representation of the body. Perhaps something like,

```java
interface TransportBodySource extends ObservableSource<ByteBuffer> {
  /**
   * Returns the total length of the body if known in advance.
   */
  Optional<Long> getContentLength();
  // Or just "long" with a negative return value for the unknown case
}

abstract class TransportBody extends Observable<ByteBuffer> implements TransportBodySource {
  public static TransportBody from(byte[]);
  public static TransportBody from(Observable<ByteBuffer>);
}
```

This would allow the HTTP transport to optimize like so,

```java
if 0 < body.getContentLength() < MAX_INMEMORY_BODY_SIZE {
  // Receive the whole body into a buffer of the given size
  // and send that as a single HTTPRequest/Response message
} else {
  // current streaming logic
}
```

Where `MAX_INMEMORY_BODY_SIZE` is the maximum size of the body which we will store in memory instead of using chunked encoding.

CC @bpon @vivekrsharma

bpon:
Fixes #197


yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpTransportDecoder.java
```diff
@@ -47,7 +47,8 @@
 @RequiredArgsConstructor
 class HttpTransportDecoder<I, O> extends ChannelInboundHandlerAdapter {
   @NonNull private final Configuration<I, O> config;
-  private final Subject<ByteBuffer> bodySubject = PublishSubject.create();
+  private TransportBodySubject bodySubject = null;
+  private long bytesRead = 0;
```
bpon: I think I need to move this logic to TransportBodySubject

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpTransportEncoder.java
```diff
@@ -71,8 +70,14 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
     @SuppressWarnings("unchecked")
     I message = (I) msg;
-    ctx.write(config.buildStartMessage(message))
-        .addListener(f -> config.getBody(message).subscribe(new Streamer(ctx, promise)));
+    TransportBody body = config.getBody(message);
+    if (body.getSize() < 0 || body.getSize() > TransportBody.MAX_BUFFERED_BODY_SIZE) {
+      ctx.write(config.buildStartMessage(message))
+          .addListener(f -> body.subscribe(new Streamer(ctx, promise)));
+    } else {
+      byte[] bodyBytes = body.accumulate().blockingGet();
```
bpon: Is this right?

bpon:
Fixes #197

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportRequestDecoder.java
```diff
@@ -42,7 +45,21 @@
...
   @Override
-  public TransportRequest build(HttpRequest request, Observable<ByteBuffer> body) {
+  public TransportBodySubject initializeTransportBody(HttpRequest request)
+      throws TransportException {
+    String contentLengthValue = request.headers().get(CONTENT_LENGTH);
+    if (contentLengthValue != null) {
+      try {
+        return TransportBodySubject.create(Long.parseLong(contentLengthValue));
+      } catch (NumberFormatException e) {
+        throw new TransportException(String.format("Invalid %s header", CONTENT_LENGTH));
```
abhinav: Might be easier to use [HttpUtil.getContentLength].

[HttpUtil.getContentLength]: https://netty.io/4.1/api/io/netty/handler/codec/http/HttpUtil.html#getContentLength(io.netty.handler.codec.http.HttpMessage,%20int)

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportRequestEncoderConfiguration.java
```diff
@@ -67,20 +69,29 @@ public boolean isInputMessage(Object message) {
...
+    httpHeaders.set(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED);
+    return httpRequest;
+  }
+
+  @Override
+  public HttpRequest buildFullMessage(TransportRequest request, byte[] body) {
+    DefaultHttpRequest httpRequest =
+        new DefaultFullHttpRequest(
+            HTTP_VERSION, HttpMethod.POST, url.getPath(), Unpooled.wrappedBuffer(body));
+    setCommonHeaders(httpRequest.headers(), request);
```
abhinav: You still need to set the content length header I think. You can probably use HttpUtil.setContentLength.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportResponseDecoder.java
```diff
@@ -46,7 +49,21 @@ public boolean isStartMessage(Object message) {
...
+  public TransportBodySubject initializeTransportBody(HttpResponse response)
+      throws TransportException {
+    String contentLengthValue = response.headers().get(CONTENT_LENGTH);
+    if (contentLengthValue != null) {
+      try {
+        return TransportBodySubject.create(Long.parseLong(contentLengthValue));
+      } catch (NumberFormatException e) {
+        throw new TransportException(String.format("Invalid %s header", CONTENT_LENGTH));
+      }
+    }
```
abhinav: HttpUtil.getContentLength

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportResponseEncoder.java
```diff
@@ -51,19 +54,31 @@ public boolean isInputMessage(Object message) {
...
+    httpHeaders.set(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED);
+    return httpResponse;
+  }
+
+  @Override
+  public HttpResponse buildFullMessage(TransportResponse response, byte[] body) {
+    DefaultHttpResponse httpResponse =
+        new DefaultFullHttpResponse(
+            HTTP_VERSION, HttpResponseStatus.OK, Unpooled.wrappedBuffer(body));
+    setCommonHeaders(httpResponse.headers(), response);
```
abhinav: set content length

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportRequestEncoderConfiguration.java
bpon: ✔️

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportResponseEncoder.java
bpon: ✔️

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportResponseDecoder.java
```diff
@@ -46,7 +48,13 @@ public boolean isStartMessage(Object message) {
   }
 
   @Override
-  public TransportResponse build(HttpResponse response, Observable<ByteBuffer> body) {
+  public TransportBodySubject initializeTransportBody(HttpResponse response)
+      throws TransportException {
+    return TransportBodySubject.create(HttpUtil.getContentLength(response, -1));
```
abhinav: FYI, both HttpRequest and HttpResponse implement HttpMessage and that's really what you need for `HttpUtil.getContentLength`. So you can actually delete this function completely and just do `HttpUtil.getContentLength` in HttpTransportDecoder. You may have to add a `extends HttpMessage` constraint on the input type in HttpTransportDecoder though.

yarpc-transport-http/src/test/java/io/yarpc/transport/http/TransportRequestDecoderTest.java
```diff
@@ -0,0 +1,147 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+
+public class TransportRequestDecoderTest {
+  @Rule public MockitoRule mockitoRule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);
+
+  @Mock ChannelHandlerContext ctx;
+  FutureSingleObserver<byte[]> resultFuture;
```
abhinav: Might be easier to use TestObserver here. You can easily create a TestObserver and subscribe it to a single/observable by calling `.test()`.

    TestObserver<byte[]> observer = res.getBody.accumulate().test();

(You can also manually instantiate and subscribe it.)

It allows you to make asserts about values posted to it, completion, etc. See http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/observers/TestObserver.html

Issue #129, opened 2017-01-17
Code coverage to at least 90%


breerly:
Please create a rule that will fail the build if coverage falls below 90%. Getting ahead of this now will ensure this project stays up to par.

PR #228, opened 2017-02-15
http: inject NioEventLoopGroup from transport

vivekrsharma:
The NioEventLoopGroup is injected by inbounds and outbounds and this can lead to a very large # of NioEventLoopGroup's very quickly. Injecting it in transport to ensure that the application has only two NioEventLoopGroup and are shared across various inbounds and outbounds.


yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpOutbound.java
```diff
@@ -66,12 +66,14 @@
...
 
-  private NioEventLoopGroup workerGroup;
+  @NonNull private NioEventLoopGroup workerGroup;
+
   private Bootstrap bootstrap;
 
   HttpOutbound(@NonNull HttpTransport transport, @NonNull URL url) {
     this.transport = transport;
     this.url = url;
+    this.workerGroup = transport.workerGroup;
```
abhinav: inject as a constructor argument or using `transport.getWorkerGroup()`

don't directly access attributes of the transport

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpTransport.java
```diff
@@ -40,6 +41,9 @@
 
   private final AtomicLifecycle lifecycle = new AtomicLifecycle();
 
+  @NonNull final NioEventLoopGroup workerGroup = new NioEventLoopGroup();
+  @NonNull final NioEventLoopGroup bossGroup = new NioEventLoopGroup();
```
abhinav: make these attributes private and possibly add getters

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpOutbound.java
vivekrsharma: Yes, agreed and fixed both.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpOutbound.java
```diff
@@ -66,12 +66,14 @@
   /** Configuration for the HttpTransportEncoder. */
   @NonNull private final TransportRequestEncoderConfiguration requestEncoderConfig;
 
-  private NioEventLoopGroup workerGroup;
+  @NonNull private NioEventLoopGroup workerGroup;
```
abhinav: final?

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -68,8 +68,10 @@
    */
   @Getter private int port;
 
-  private NioEventLoopGroup bossGroup;
-  private NioEventLoopGroup workerGroup;
+  @NonNull private NioEventLoopGroup bossGroup;
+
+  @NonNull private NioEventLoopGroup workerGroup;
```
abhinav: final?

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpOutbound.java
```diff
@@ -97,9 +99,7 @@ public void start() {
   }
 
   private void doStart() {
-    // TODO: Separate event loops for each outbound/inbound seems wrong.
-    // The Transport should inject this.
-    this.workerGroup = new NioEventLoopGroup();
+
```
abhinav: nit: delete empty line

breerly:
What's with the coveralls failing?

breerly:
@vivekrsharma i understand what you're doing here, just not why you are doing it.

Can you update the PR description to include the motivation for the change, so that others can more easily review it?

bpon:
@breerly I think the coveralls threshold is set to 90%.  We're still working on it (#129).

breerly:
@bpon seems like we should set it to 60% for now so that builds pass, and then keep raising it as the coverage increases... or is there some advantage to this approach?

vivekrsharma:
@breerly So spoke to abg and looks like we would want the coverage to be 90% and just ignore the build failure due to incomplete coverage for now.

Issue #220, opened 2017-02-10
Client call reply objects


abhinav:
With #219, we've decided on the API for client calls. We also need to figure out how a caller receives response metadata like response headers.

The offline discussion lead to the idea of returning a `Reply` object rather than just the result for all calls. So instead of returning `CompletionStage<Result>`, we'll return `CompletionStage<Reply<Result>>` where `Reply` is similar to,

```java
interface Reply<T> {
  Headers getHeaders();
  T getBody();
}
```

Usage:

```java
client.getValue(key).call(ctx)
  .whenComplete((reply, ex) -> {
    // ..
    
    System.out.println(reply.getHeaders());
    System.out.println(reply.getBody());
  });
```

Issue #219, opened 2017-02-10
Call-based Client APIs


abhinav:
Documenting the results of an offline discussion here:

We were trying to figure out how we want users to specify request metadata at call sites. Main options that were considered were:

- Put everything on the context. This is really confusing because incoming request metadata is also on the context. You end up having to separate incoming and outgoing contexts into separate types and having weird construction semantics.
- Variadic arguments
- Call objects

We decided that call objects are the way to go. Every client API call, instead of returning a `CompletionStage<Result>` will return a `Call<Result>` where `Call` is defined as:

```java
interface Call<T> {
  Call<T> header(String key, String value);
  Call<T> shardKey(String shardKey);
  // ...
  
  CompletionStage<Result> call(Context ctx);
}
```

This will be used like so,

```java
client
  .getValue(key)
  .header("foo", "bar")
  .routingDelegate("rd")
  .call(ctx)
  .whenComplete((result, ex) -> {
    // ...
  });
```

Issue #221, opened 2017-02-10
Server-side response headers


abhinav:
Follow up discussion to #219 and #220: To return response headers from server-side handlers, we'll use the Context object.

```java
interface Context {
  void writeResponseHeader(String key, String value);
  // ...
}
```

The encoding will be responsible for tracking the headers written to the context and sending them over the wire. Pseudo-code;

```java
private UnaryRequestHandler handler;

Single<TransportResponse> handle(TransportRequest request) {
  Context ctx = Context.fromRequest(request);
  I data = decode(request.getBody());
  return handler.handleUnary(ctx, data)
    .flatMap(o -> TransportResponse.from(ctx, o));
}
```

PR #229, opened 2017-02-15
core: Implement Call and Reply

bpon:
#219 #220


bpon:
@abhinav @vivekrsharma

yarpc-core/src/main/java/io/yarpc/core/transport/Call.java
```diff
@@ -0,0 +1,54 @@
...
+  Call<T> header(String key, String value);
+
+  CompletionStage<T> call(Context ctx) throws TransportException;
+
+  default <U> Call<U> map(Function<? super T, ? extends U> func) {
+    return new MappedCall<>(this, func);
+  }
+
+  @RequiredArgsConstructor
+  class MappedCall<T, U> implements Call<U> {
```
abhinav: Move to package-local top-level class?

Also should we put this class and its implementations in their own package if
we're planning on having a bunch of private implementations?

yarpc-core/src/main/java/io/yarpc/core/transport/DefaultCall.java
```diff
@@ -0,0 +1,51 @@
...
+    return this;
+  }
+
+  @Override
+  public final CompletionStage<T> call(Context ctx) throws TransportException {
+    return doCall(
+        Context.builder()
+            .headers(
+                ImmutableMap.<String, String>builder()
+                    .putAll(ctx.getHeaders())
```
abhinav: Is this how we want to store this information for outgoing calls?

I was imagining that this would be the part of the system where we build the
TransportRequest from the parameters stored in the Call.

Like, the encoding implements the part of `Call` where the request body
becomes bytes for the outgoing `TransportRequest`.

yarpc-core/src/main/java/io/yarpc/core/transport/UnaryCall.java
```diff
@@ -0,0 +1,48 @@
...
+  protected final CompletionStage<Reply<T>> doCall(Context ctx) throws TransportException {
+    TransportRequest req = buildRequest(ctx);
+    return clientConfig
+        .getUnaryOutbound()
+        .callUnary(req)
+        .flatMap(this::handleResponse)
+        .to(CompletionStages::fromSingle);
+  }
+
+  protected abstract TransportRequest buildRequest(Context ctx);
```
abhinav: Yeah, something like this but not using the `Context` to store the outgoing
call information. Reason being that `Context` would be used only for incoming
metadata. See also #221.

yarpc-core/src/main/java/io/yarpc/core/transport/Call.java
bpon: I was noticing that the `io.yarpc.core.transport` package is getting pretty fat.  Let's discuss if we want to reorganize it after this PR?

yarpc-encoding-raw/src/main/java/io/yarpc/encoding/raw/RawClient.java
```diff
@@ -22,66 +22,74 @@
...
+      public CompletionStage<Reply<byte[]>> call(Context ctx) throws TransportException {
+        TransportRequest req = buildTransportRequest(this, ctx, procedure, reqBody);
+        return clientConfig
+            .getUnaryOutbound()
+            .callUnary(req)
+            .flatMap(
+                res ->
+                    res.getBody()
+                        .accumulate()
+                        .map(resBody -> new DefaultReply<>(res.getHeaders(), resBody)))
```
abhinav: Readability nit: too many levels; please add a function somewhere

yarpc-encoding-thrift/src/main/java/io/yarpc/encoding/thrift/ThriftClient.java
```diff
@@ -46,58 +49,55 @@
...
+            .getUnaryOutbound()
+            .callUnary(req)
+            .flatMap(
+                res ->
+                    res.getBody()
+                        .accumulate()
+                        .map(
+                            resBody -> {
+                              R resData = resClass.newInstance();
+                              TDeserializer deserializer =
```
abhinav: Same

PR #231, opened 2017-02-16
encoding: Add methods for constructing procedures

bpon:
For each encoding, add methods to construct unary and oneway procedures from request handlers.


yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/JsonEncoding.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Creates a new unary procedure with JSON encoding.
+   *
+   * @param name the procedure name
+   * @param service the service name
+   * @param requestHandler a request handler
+   * @param inputClass the json-serializable input type
+   */
+  public static <I, O> Procedure newUnaryProcedure(
+      String name, String service, UnaryRequestHandler<I, O> requestHandler, Class<I> inputClass) {
```
abhinav: Drop the service parameter for now. We'll want to use the service name that the Dispatcher is using.

yarpc-encoding-thrift/src/main/java/io/yarpc/encoding/thrift/ThriftEncoding.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Creates a new unary procedure with Thrift encoding.
+   *
+   * @param name the procedure name
+   * @param requestHandler a request handler
+   * @param inputClass the thrift-serializable input type
+   */
+  public static <I extends TBase, O extends TBase> Procedure newUnaryProcedure(
+      String name, UnaryRequestHandler<I, O> requestHandler, Class<I> inputClass) {
```
abhinav: We'll have to figure out how we want generated code to expose Procedures from user handlers as well

PR #232, opened 2017-02-17
core: Default procedure service name to dispatcher name

bpon:
Use the Dispatcher's name as the default service name when registering procedures.


yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
```diff
@@ -73,11 +73,20 @@
 
   @NonNull final OutboundMiddleware outboundMiddleware;
 
-  @NonNull private final RouteTable routeTable = new MapRouteTable();
+  private final RouteTable routeTable;
```
abhinav: why drop the `@NonNull`?

yarpc-core/src/main/java/io/yarpc/core/transport/MapRouteTable.java
```diff
@@ -60,8 +66,8 @@ public HandlerSpec choose(TransportRequest request) throws TransportException {
 
     @NonNull private final String procedure;
 
-    public ServiceProcedure(Procedure procedure) {
-      this(procedure.getService(), procedure.getName());
+    public ServiceProcedure(String service, Procedure procedure) {
+      this(service, procedure.getName());
```
abhinav: This really just stores the procedure name. Any reason to not make that the argument?

yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
bpon: oops

PR #237, opened 2017-02-21
Integration tests with raw and json encoding

bpon:
Integration tests with raw and json encoding over http.

I'm also working on tests with thrift encoding, but I discovered that I need to add an additional gradle task to generate the YARPC client code prior to building test sources.  This will be in a separate PR.


yarpc-integration-tests/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -21,119 +21,123 @@
...
-
-  private static Outbound setupHttpClient() throws IOException, InterruptedException {
-    return new HTTPOutbound(Inet4Address.getLocalHost().getHostAddress(), httpPort);
+  @BeforeClass
+  public static void startDispatcher() throws Exception {
+    HttpTransport httpTransport = new HttpTransport();
+    HttpInbound inbound = httpTransport.inboundBuilder().port(TestUtil.findRandomPort()).build();
+    serverDispatcher = Dispatcher.builder().name(service).inbound(inbound).build();
+    serverDispatcher.register(
+        ImmutableList.of(
```
abhinav: For later: We may want to allow register to take a variadic number of procedures.

yarpc-integration-tests/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -21,119 +21,123 @@
...
+            JsonEncoding.newUnaryProcedure(echoProcedure, new EchoHandler(), KeyValue.class),
+            JsonEncoding.newOnewayProcedure(
+                returnNullProcedure, new ReturnNullHandler(), KeyValue.class)));
+    serverDispatcher.start();
+    HttpOutbound outbound =
+        httpTransport.outboundBuilder().url(TestUtil.getLocalhostUrl(inbound.getPort())).build();
+    clientDispatcher =
+        Dispatcher.builder()
+            .name(service + "-client")
+            .outbound(service, (UnaryOutbound) outbound)
```
abhinav: Should we rename these functions if overloading is going to be a problem?

yarpc-integration-tests/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -21,119 +21,123 @@
...
-            .procedure(endpoint)
-            .timeout(Duration.ofMillis(1000))
-            .build();
-    ListenableFuture<Response<TestResult>> future = client.call(request, TestResult.class);
-    Response<TestResult> response = future.get();
-    Assert.assertEquals(response.getBody().getA(), respString);
+  private static class EchoHandler implements UnaryRequestHandler<KeyValue, KeyValue> {
+    @Override
+    public CompletionStage<Reply<KeyValue>> handleUnary(Context ctx, KeyValue data) {
+      return CompletableFuture.completedFuture(new DefaultReply<>(new Headers(), data));
```
abhinav: Perhaps add a TODO to also return the request headers in the response

yarpc-integration-tests/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -21,119 +21,123 @@
...
+    }
   }
 
-  private Channel createChannel(Outbound outbound) {
-    return Channel.builder().caller("client").service(rpcName).outbound(outbound).build();
+  private static class ReturnNullHandler implements OnewayRequestHandler<KeyValue> {
+    @Override
+    public CompletionStage<Void> handleOneway(Context ctx, KeyValue data) {
+      return TestUtil.nullFuture();
```
abhinav: NoopOnewayHandler?

yarpc-integration-tests/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
bpon: Yeah, lemme create an issue to fix that later.

yarpc-integration-tests/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
bpon: 👍  but using "Nop" instead of "Noop" for consistency with other classes (I personally prefer "Noop" but I started using "Nop", since that's what yarpc-go uses).

PR #244, opened 2017-02-23
core: Reorganize code in transport package

bpon:


yarpc-core/src/main/java/io/yarpc/core/internal/AtomicLifecycle.java
```diff
@@ -19,7 +19,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package io.yarpc.core.transport;
+package io.yarpc.core.internal;
```
vivekrsharma: was this supposed to be io.yarpc.core.transport.internal?

bpon: That's overkill.  We just need one internal package at the top level package of the project.  Also, we were gonna move Lifecycle out of the transport package anyway.

vivekrsharma: Okay

Issue #247, opened 2017-02-23
thrift: Client and Server interfaces


abhinav:
Currently the Thrift code we generate exposes the `*_args` and `*_result` objects. This isn't how we want the user experience to be.

Given,

```
service KeyValue {
  void setValue(1: Key key, 2: Value value)
  Value getValue(1: Key key)
    throws (1: ValueDoesNotExist doesNotExist)
}
```

As a service author, I should only implement,

```
class KeyValueImpl implements KeyValue.Handler {
  public CompletionStage<Reply<Void>>
    setValue(Context ctx, Key key, Value value) {
      // ...
  }
  
  public CompletionStage<Reply<Value>> getValue(Context ctx, Key key) {
    return CompletionStage.failedFuture(new ValueDoesNotExist(...));
  }
}
```

How this is registered against the Dispatcher is up for discussion but some ways are,

```
KeyValue.register(dispatcher, new KeyValueImpl());
dispathcer.register(new KeyValueImpl());
dispatcher.register(KeyValue.procedures(new KeyValueImpl());
```

In the background, we'll want to adapt the `KeyValueHandler` into a list of `Procedure`s which hold `UnaryRequestHandler<getValue_args, getValue_result>` and similar.

On the client side, we already expose arguments rather than the `*_args` structs but we still return the `*_result` object which users have to manually interpret the exceptions from. We should change that to always return just the return value (or `Void`) and fail the future with the corresponding exception if an exception was received.

---

To implement this, our yarpc-encoding-thrift code will basically be generic and parameterized over the `_args` and `_result` types but the generated code will adapt that to the user-facing interfaces. This is similar to the approach employed by YARPC Go.

bpon:
We will have to discuss the client side further.  There are a few ways we could handle the response.

1. The current way is to generate a unique `*Reply` type for each procedure that throws thrift exceptions when unwrapping the response data.

```java
client.getValue("foo")
  .call(context)
  .thenAccept(/*GetValueReply*/ reply -> {
    Headers headers = reply.getHeaders();
    // Do something with headers
    try {
      String value = reply.get();
      // Do something with value
    } catch (ValueNotFound exception) {
      // Handle exception
    }
  });
```

Pros:
- Forces user to handle exceptions
- Allows access to headers for both successful and exceptional responses

Cons:
- Requires a unique (but generated) reply type for each procedure
- Users need to call `reply.get()` even for void responses

2. Another option is to handle thrift exceptions with a CompletionStage.

[...]

PR #248, opened 2017-02-24
thrift-generator: Generate handler interface and procedures

bpon:
This addresses #247 by generating a handler interface and procedures for thrift services.


yarpc-thrift-generator/src/main/resources/templates/yarpc/common.stg
```diff
@@ -4,26 +4,44 @@ package <context.javaPackage>;
...
 >>
 
-_method(method) ::= <<
-public io.yarpc.core.Call\<<if(method.oneway)>io.yarpc.core.transport.Ack<else><_reply(method)><endif>\> <method.javaName><_paramsWithContext(method.parameters)> {
-    <_args(method)> args = new <_args(method)><_callParams(method.parameters)>;
+_procedure(method) ::= <<
+io.yarpc.encoding.thrift.ThriftEncoding.new<if(method.oneway)>Oneway<else>Unary<endif>Procedure("<method.javaName>", new <_handler(method)>(handler), <_args(method)>.class)
+>>
+
+_clientMethod(method) ::= <<
```
abhinav: Does this format allow for comments in the template that don't get rendered? If yes, we should add some information above these sections later.

yarpc-thrift-generator/src/main/resources/templates/yarpc/common.stg
```diff
@@ -87,18 +85,109 @@ public static final class <_reply(method)> {
...
+
+    @Override
+    public java.util.concurrent.CompletionStage\<io.yarpc.core.Reply\<<_result(method)>\>\> handleUnary(io.yarpc.core.Context ctx, <_args(method)> args) {
+        try {
+            args.validate();
+        } catch (org.apache.thrift.TException e) {
+            throw new io.yarpc.encoding.thrift.ThriftEncodingException(e);
+        }
+        return handler.<method.javaName>(ctx<if(method.parameters)>, <method.parameters: {param |args.<param.javaGetterName>()}; separator=", "><endif>)
+            .handle((reply, ex) -> {
```
abhinav: Should we handle the case where the handle function returns null?

bpon: #251

yarpc-thrift-generator/src/main/java/io/yarpc/thrift/generator/YarpcMethodContext.java
```diff
@@ -44,12 +45,20 @@ static YarpcMethodContext fromSwiftMethodContext(
...
         .parameters(methodContext.getParameters())
         .exceptions(methodContext.getExceptions())
         .build();
   }
+
+  private static String capitalize(String str) {
+    if (str.length() > 1) {
+      return str.substring(0, 1).toUpperCase() + str.substring(1);
+    }
+    return str.toUpperCase();
```
bpon: length should be never zero, but even if it is, it'll be handled by `toUpperCase`.

abhinav: That's what I meant. It'll never be zero but you have special handling for it. If it really is zero, just returning `str` is the same as `str.toUpperCase()`.

bpon: This line is for handling length == 1.

abhinav: facepalm.jpg

My bad

yarpc-thrift-generator/src/main/resources/templates/yarpc/common.stg
bpon: Comments added

PR #253, opened 2017-02-28
dispatcher: Add middleware to inbounds and outbounds

bpon:
Adds middleware to outbounds when the Dispatcher is created.  Adds middleware to inbounds when procedures are registered.


yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
```diff
@@ -266,57 +301,112 @@ public DispatcherBuilder outboundMiddleware(OnewayOutboundMiddleware onewayOutbo
...
+                  outboundConfig ->
+                      SimpleClientConfig.builder()
+                          .caller(name)
+                          .service(outboundConfig.getKey())
+                          .unaryOutbound(
+                              unaryOutboundWithMiddleware(
+                                  outboundConfig.getValue().getUnaryOutbound(),
+                                  unaryOutboundMiddlewareChain))
+                          .onewayOutbound(
+                              onewayOutboundWithMiddleware(
```
abhinav: My eyes!

Less nesting please. A function or two never hurt anyone. Pulling this into a function will allow you to move logic from unaryOutboundMiddleware across multiple statements.

```
SimpleClientConfigBuilder builder = ...;
if (cfg.getUnaryOutbound() != null) {
  builder = builder.unaryOutbound(...);
}
...
return builder.build();
```

yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
```diff
@@ -266,57 +301,112 @@ public DispatcherBuilder outboundMiddleware(OnewayOutboundMiddleware onewayOutbo
...
+    private static OnewayOutbound onewayOutboundWithMiddleware(
+        OnewayOutbound outbound, OnewayOutboundMiddleware middleware) {
+      if (outbound == null) {
+        return null;
+      }
+      return new OnewayOutboundWithMiddleware(outbound, middleware);
+    }
+
+    @RequiredArgsConstructor
+    private static class UnaryInboundMiddlewareChain implements UnaryInboundMiddleware {
```
abhinav: Technically, this is a link in the chain, not the chain itself.

yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
```diff
@@ -266,57 +301,112 @@ public DispatcherBuilder outboundMiddleware(OnewayOutboundMiddleware onewayOutbo
...
+      private final UnaryInboundMiddleware prev;
+      private final UnaryInboundMiddleware next;
+
+      @Override
+      public Single<TransportResponse> handleUnary(
+          TransportRequest req, UnaryTransportHandler handler) throws TransportException {
+        return prev.handleUnary(req, new UnaryTransportHandlerWithMiddleware(handler, next));
```
abhinav: This will result in a lot of layers of wrapping for each middleware in the
list. We can avoid all this by tracking per-call state in an iterator.


```java
@RequiredArgsConstructor
class UnaryInboundMiddlewareChain implements UnaryInboundMiddleware {
  @NonNull private final Iterable<UnaryInboundMiddleware> middlewares;

  @Override
  public Single<TransportResponse> handleUnary(
    TransportRequest req, UnaryTransportHandler handler
  ) throws TransportException {
    return new Handler(middlewares.iterator(), handler).handleUnary(req);
  }
  
  @RequiredArgsConstructor
  private final class Handler implements TransportHandler {
    @NonNull private final Iterator<UnaryInboundMiddleware> middlewares;
    @NonNull private final UnaryTransportHandler handler;
  
    @Override
    public Single<TransportResponse> handleUnary(TransportRequest req)
        throws TransportException {
      if (!iter.hasNext()) {
        return handler.handleUnary(req);
      }
      return iter.next().handleUnary(req, this);
    }
  }
}
```

(Aside: Are we actually throwing TransportExceptions outside the Single
anymore?)

bpon: Ah didn't think of that.  Sounds good to me.

bpon: Yeah it looks like we're not throwing TransportExceptions here anymore.  I'll clean this up later.

yarpc-core/src/main/java/io/yarpc/core/Dispatcher.java
```diff
@@ -266,57 +297,54 @@ public DispatcherBuilder outboundMiddleware(OnewayOutboundMiddleware onewayOutbo
...
+
+    private static UnaryOutbound unaryOutboundWithMiddleware(
+        UnaryOutbound outbound, UnaryOutboundMiddleware middleware) {
+      if (outbound == null) {
+        return null;
+      }
+      return new UnaryOutboundWithMiddleware(outbound, middleware);
+    }
+
+    private static OnewayOutbound onewayOutboundWithMiddleware(
```
abhinav: Nit: You can get rid of this and the other function by having two simple if statements in the `outboundClientConfig` function. IMO for something this simple, an if statement is more readable than a helper function.

yarpc-core/src/main/java/io/yarpc/core/transport/middleware/OutboundWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package io.yarpc.core.transport.middleware;
+
+import io.yarpc.core.transport.Outbound;
+import io.yarpc.core.transport.Transport;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+abstract class OutboundWrapper implements Outbound {
```
abhinav: I don't feel strongly about this but IIRC, classes like this one are usually called `ForwardingFoo` or `FooDelegate`.

yarpc-core/src/main/java/io/yarpc/core/transport/middleware/OutboundWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+  @NonNull private final Outbound outbound;
```
abhinav: Does it make sense to have a protected getter for this so that implementors can access the underlying outbound without maintaining a reference to them by hand?

yarpc-core/src/main/java/io/yarpc/core/transport/middleware/OutboundWrapper.java
bpon: I don't recall seeing those, but I've seen these: https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletResponseWrapper.html
http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/methods/HttpRequestWrapper.html

PR #258, opened 2017-03-01
Handle response headers

bpon:


yarpc-thrift-generator/src/main/resources/templates/yarpc/common.stg
```diff
@@ -181,14 +184,17 @@ if (result.isSet<exception.type>()) {
...
 <if(method.exceptions)>
 if (ex instanceof org.apache.thrift.TException) {
-    <exceptions: {exception |<_checkAndSetException(exception)>}; separator=" else "> else {
+    <method.exceptions: {exception |<_checkAndSetException(exception)>}; separator=" else "> else {
         throw new io.yarpc.encoding.thrift.ThriftEncodingException(ex);
     }
     <! TODO Can we still produce meaningful headers? !>
-    return new io.yarpc.core.DefaultReply\<\>(new io.yarpc.core.transport.Headers(), result);
+    return io.yarpc.core.DefaultReply.\<<_result(method)>\>builder()
+        .headers(new io.yarpc.core.transport.Headers())
```
abhinav: shouldn't this hold the TransportResponse headers?

bpon: At the moment, handlers can't throw exceptions and provide response headers at the same time.  We'll have to figure out how to do this.

PR #264, opened 2017-03-03
http: Handle transport exceptions

bpon:
This provides transport-level exception handling.

Still TODO:
- [x] Define specific exception types
- [x] Return different HTTP status codes


yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpTransportDecoder.java
```diff
@@ -54,6 +54,7 @@
   public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
     if (!(msg instanceof HttpObject)) {
       ctx.fireChannelRead(msg);
+      return;
```
abhinav: oo that must've been an annoying bug

yarpc-transport-http/src/main/java/io/yarpc/transport/http/Channels.java
```diff
@@ -63,4 +74,25 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+      observer.onError(cause);
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+      if (exceptionClass.isInstance(msg)) {
+        observer.onError(exceptionClass.cast(msg));
```
bpon: You're right... I actually removed this class in my local branch yesterday.  Gonna update this PR today.

yarpc-integration-tests/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -109,11 +116,27 @@ public void testUnaryCall() throws Exception {
...
+    try {
+      client
+          .unary(replyNullProcedure, data, KeyValue.class)
+          .call(Context.builder().build())
+          .toCompletableFuture()
+          .get();
+      fail("An exception should be thrown when the reply is null");
+    } catch (ExecutionException e) {
+      // TODO More specific exception type: UnexpectedException?
+      assertTrue(e.getCause() instanceof TransportException);
```
bpon: Because the exception being thrown is an ExecutionException.  I suppose I could throw `e.getCause()` instead.

bpon:
Looks like a huge PR, but the majority of it is new exception types and tests.

yarpc-core/src/main/java/io/yarpc/core/transport/exceptions/RequestEncodingException.java
```diff
@@ -0,0 +1,64 @@
...
+ * THE SOFTWARE.
+ */
+package io.yarpc.core.transport.exceptions;
+
+import io.yarpc.core.transport.TransportRequest;
+
+/** Error thrown when trying to decode a request body. */
+public class RequestEncodingException extends BadRequestException {
+  public RequestEncodingException(
+      String service, String procedure, String caller, String encoding, String message) {
```
bpon: Good idea... lemme do that in the next PR along with asserting the exception messages in tests.

PR #267, opened 2017-03-08
crossdock/oneway: Enable oneway raw

vivekrsharma:
This integrates the cross language tests with Go.
Currently uses raw encoding.


Makefile
```diff
@@ -21,8 +21,8 @@ coverage:
 	./gradlew jacocoRootReport
 
 dockclean:
-	docker-compose kill
-	docker-compose rm --force
+	docker-compose kill java
+	docker-compose rm -f java
```
abhinav: IMO you should change this back. `clean` should mean clean everything, including dependencies.

Feel free to drop the dependency on clean in the `crossdock` target and/or clean-only Java there explicitly.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/oneway/OnewayHandler.java
```diff
@@ -0,0 +1,68 @@
...
+import io.yarpc.core.transport.Headers;
+import io.yarpc.core.transport.TransportBody;
+import io.yarpc.core.transport.TransportRequest;
+import io.yarpc.crossdock.server.OnewaySubject;
+import io.yarpc.encoding.raw.RawEncoding;
+import io.yarpc.transport.http.HttpOutbound;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+/** OnewayHandler is a mock for oneway. */
```
abhinav: it's not a mock

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/oneway/OnewayHandler.java
```diff
@@ -0,0 +1,68 @@
...
+public class OnewayHandler implements OnewayRequestHandler<byte[]> {
+  public static final String CALL_BACK_ADDR = "callBackAddr";
+  private static final String ONEWWAY_SERVER = "oneway-server";
```
abhinav: typo: `ONEWAY_SERVER`

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/oneway/OnewayHandler.java
```diff
@@ -0,0 +1,68 @@
...
+  private static final String ONEWAY_CLIENT = "oneway-client";
+  private static final String CALLBACK = "call-back";
+
+  @SuppressFBWarnings("NP_NONNULL_PARAM_VIOLATION")
+  @Override
+  public CompletionStage<Void> handleOneway(Context ctx, byte[] body) {
+    callHome(ctx, ctx.getHeaders().get(CALL_BACK_ADDR), body);
+    return CompletableFuture.completedFuture(null);
```
abhinav: You should make callHome return the `Single<Ack>` you get by calling `callOneway` and build a CompletableFuture from that instead of returning this one always. That way if the outbound call fails immediately, this doesn't count as a success.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/server/OnewaySubject.java
```diff
@@ -0,0 +1,53 @@
...
+import io.yarpc.transport.http.HttpTransport;
+
+public class OnewaySubject implements Subject {
+  private static final int HTTP_PORT = 8084;
+  public static final String SERVICE_NAME = "oneway-server";
+  private Dispatcher serverDispatcher;
+  public static final HttpTransport httpTransport = new HttpTransport();
+
+  @Override
+  /** Starts the transport and server dispatcher for Yarpc. */
```
abhinav: docs go above the annotations

yarpc-crossdock/src/main/java/io/yarpc/crossdock/server/OnewaySubject.java
```diff
@@ -0,0 +1,53 @@
...
+import io.yarpc.crossdock.behavior.oneway.OnewayHandler;
+import io.yarpc.encoding.raw.RawEncoding;
+import io.yarpc.transport.http.HttpInbound;
+import io.yarpc.transport.http.HttpTransport;
+
+public class OnewaySubject implements Subject {
+  private static final int HTTP_PORT = 8084;
+  public static final String SERVICE_NAME = "oneway-server";
+  private Dispatcher serverDispatcher;
+  public static final HttpTransport httpTransport = new HttpTransport();
```
abhinav: why is the transport static but not the dispatcher? IMO none of this should be static.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/server/OnewaySubject.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Override
+  /** Starts the transport and server dispatcher for Yarpc. */
+  public void start() {
+    HttpInbound inbound = httpTransport.inboundBuilder().port(HTTP_PORT).build();
+    serverDispatcher = Dispatcher.builder().name(SERVICE_NAME).inbound(inbound).build();
+    serverDispatcher.register(RawEncoding.newOnewayProcedure("echo/raw", new OnewayHandler()));
```
abhinav: Preferably, instantiate everything in the constructor and start it in start.

So this start ends up being just `serverDispatcher.start()`.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -176,6 +178,11 @@ protected void channelRead0(ChannelHandlerContext ctx, TransportRequest request)
           handlerSpec.getUnary().handleUnary(request).subscribe(new UnaryResponseHandler(ctx));
           break;
         case Oneway:
+          TransportResponse response =
+              DefaultTransportResponse.builder()
+                  .body(TransportBody.fromByteArray(new byte[0]))
+                  .build();
```
abhinav: This seems like a change separate from the rest of this PR. Please pull this into a different PR.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -217,16 +224,12 @@ public void onSubscribe(Disposable disposable) {
 
     @Override
     public void onComplete() {
-      // TODO close channel as soon as request is done reading rather than after handling the
-      // response
-      ctx.close();
+      // nothing to do
```
abhinav: If we're not doing anything for all other cases, it may be preferable to use [doOnError].

  [doOnError]: http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Completable.html#doOnError(io.reactivex.functions.Consumer)

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -217,16 +224,12 @@ public void onSubscribe(Disposable disposable) {
...
     }
 
     @Override
     public void onError(Throwable cause) {
-      // TODO send back 500 error instead
-      // TODO log failure
-      ctx.close();
+      // TODO: handle errors
```
abhinav: We don't want to fire exception caught after we've closed the channel (which is possible if we finished writing to it before we finished processing the oneway request). We should log the error instead.

Makefile
vivekrsharma: Fixed.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
vivekrsharma: Pulled out the HttpInbound.java in another PR

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/oneway/RawOnewayHandler.java
```diff
@@ -0,0 +1,71 @@
...
+  public CompletionStage<Void> handleOneway(Context ctx, byte[] body) {
+    return callHome(ctx, ctx.getHeaders().get(CALL_BACK_ADDR), body)
+        .to(CompletionStages::fromSingle)
+        .thenApply(ack -> null);
+  }
+
+  private Single<Ack> callHome(Context ctx, String callBackAddr, byte[] body) {
+    HttpTransport httpTransport = new HttpTransport();
+    HttpOutbound outbound = httpTransport.outboundBuilder().url("http://" + callBackAddr).build();
+    outbound.start();
```
bombela: you never stop this outbound

docker-compose.yml
```diff
@@ -26,11 +28,12 @@ services:
             - AXIS_HTTPSERVER=java  # TODO: go
             - AXIS_CLIENT_ONEWAY=go
```
vivekrsharma: we don't have client for java yet.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/oneway/RawOnewayHandler.java
```diff
@@ -0,0 +1,76 @@
...
+              .caller(ONEWAY_SERVER)
+              .service(ONEWAY_CLIENT)
+              .procedure(CALLBACK)
+              .encoding(RawEncoding.ENCODING)
+              .body(TransportBody.fromByteArray(body))
+              .build();
+
+      return outbound.callOneway(transportRequest);
+    } finally {
+      outbound.stop();
```
vivekrsharma: Yes, that's async java :) Done.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/oneway/RawOnewayHandler.java
```diff
@@ -0,0 +1,71 @@
...
+        DefaultTransportRequest.builder()
+            .headers(new Headers(ctx.getHeaders()))
+            .caller(ONEWAY_SERVER)
+            .service(ONEWAY_CLIENT)
+            .procedure(CALLBACK)
+            .encoding(RawEncoding.ENCODING)
+            .body(TransportBody.fromByteArray(body))
+            .build();
+
+    return outbound.callOneway(transportRequest).doFinally(() -> outbound.stop());
```
abhinav: nit: I think you can do `outbound::stop` here.

    doFinally(outbound::stop)

yarpc-crossdock/src/main/java/io/yarpc/crossdock/server/OnewaySubject.java
```diff
@@ -0,0 +1,60 @@
...
+  /** Starts the transport and server dispatcher for Yarpc. */
+  @Override
+  public void start() {
+    serverDispatcher.start();
+  }
+
+  /** Stops the server dispatcher. */
+  @Override
+  public void stop() {
+    if (serverDispatcher == null) {
```
abhinav: the server is never null

PR #268, opened 2017-03-08
http/inbound: Close ChannelHandlerContext immediately in oneway

vivekrsharma:
This ensures that channelHandlerContext is closed immediately after the
oneway request is received. Addiitonally also changes the CompletableObserver to
DisposableCompletableObserver to handle errors.


yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -207,27 +215,21 @@ public void onError(Throwable cause) {
...
     @Override
     public void onComplete() {
-      // TODO close channel as soon as request is done reading rather than after handling the
-      // response
-      ctx.close();
+
     }
 
     @Override
-    public void onError(Throwable cause) {
```
vivekrsharma: I couldn't find an interface that only had doOnError.

abhinav: I linked to it in the original comment: http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Completable.html#doOnError(io.reactivex.functions.Consumer)

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -207,27 +215,21 @@ public void onError(Throwable cause) {
...
-      // TODO send back 500 error instead
-      // TODO log failure
-      ctx.close();
-      ctx.fireExceptionCaught(cause);
+    public void onError(Throwable e) {
+      log.error(e.toString());
```
abhinav: Don't `toString` exceptions. The last argument to SLF4J loggers can just be an exception and the stack trace will be logged.

PR #278, opened 2017-03-14
crossdock: Initial behaviors framework

vivekrsharma:
This sets up the framework for outbound and enables outbounds for raw encoding.

So far Crossdock tests for YARPC Java have been incoming-only. The behaviors
are implemented in other languages and Java only provides the test subjects.
This change sets up the framework to allow writing Crossdock behaviors in
YARPC Java.


yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,14 +32,26 @@
...
 import java.util.List;
+import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class CrossdockServerInboundHandler extends SimpleChannelInboundHandler<FullHttpRequest> {
-  private static final String SKIPPED = "skipped";
-  private static final String PASSED = "passed";
-  private static final String FAILED = "failed";
+  private static final Map<String, Behavior> behaviorClient = new HashMap<String, Behavior>();
```
abhinav: let's not make this static

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,14 +32,26 @@
...
+  private static final String RAW_BEHAVIOR = "raw";
+  private static final String BEHAVIOR = "behavior";
+  private static final String SERVER = "server";
+  private static final String TRANSPORT = "transport";
```
abhinav: I don't like making every string a class-level constant. If it's used more than once, yes, but otherwise leave it as-is. It hurts readability more than it helps.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
abhinav: also, this isn't a client. this is a map of behaviors.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/CrossdockParam.java
```diff
@@ -0,0 +1,33 @@
...
+package io.yarpc.crossdock.client;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public class CrossdockParam {
+  String encoding;
+  String server;
```
abhinav: Drop this struct. You can't contain all the parameters of all the behaviors in a struct. You'll want to pass some sort of object to the behaviors. Each behavior has its own parameters and these will be accessed using a `.get` method. Something like

    run(CrossdockParam params) {
      String encoding = param.get("encoding");

abhinav: Also, given that multiple parameters are contained in the object, we would want to name it CrossdockParams or CrossdockRequest instead of CrossdockParam.

I'm leaning towards CrossdockTestRequest with a `getParam` method so that we can attach more information to the request.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -45,27 +60,34 @@ protected void channelRead0(
...
-            + server
-            + " transport:"
-            + transport
-            + " encoding:"
-            + encoding);
 
-    // TODO: Make a client call and respond with appropriate message
+    Behavior br = behaviorClient.get(behavior);
+    CrossdockTestResponse testResponse =
+        new CrossdockTestResponse(CrossdockUtils.SKIPPED, CrossdockUtils.SKIPPED);
```
abhinav: Just write an empty array in the response instead

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -45,27 +60,34 @@ protected void channelRead0(
...
+    // TODO: Throw an UnsupportedOperationException once json, thrift are fixed.
+    if (br != null) {
+      CrossdockParam param =
+          CrossdockParam.builder().encoding(encoding).server(server).transport(transport).build();
+      testResponse = br.run(param);
```
abhinav: This isn't scalable to multiple behaviors. See comment in CrossdockParam.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/raw/EchoRawClient.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.concurrent.TimeUnit;
+
+public class EchoRawClient implements Behavior {
+  private static final String SERVICE = "yarpc-test";
+  private static final String ECHO_PROCEDURE = "echo/raw";
+
+  @Override
+  public CrossdockTestResponse run(CrossdockParam param) throws Exception {
+    Dispatcher clientDispatcher = DispatcherBuilder.createDispatcher(param.getServer());
+    clientDispatcher.start();
```
abhinav: This is never stopped. Do a try-finally perhaps.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/Behavior.java
```diff
@@ -0,0 +1,28 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.crossdock.client;
+
+import io.yarpc.crossdock.CrossdockTestResponse;
+
+public interface Behavior {
+  public CrossdockTestResponse run(CrossdockParam param) throws Exception;
```
abhinav: Consider renaming CrossdockTestResponse to CrossdockResponse (and similarly, CrossdockTestRequest proposed elsewhere to CrossdockRequest)

abhinav: Also this definition of CrossdockTestResponse is wrong. One behavior can
actually have multiple status updates.

You'll want to be able to provide zero or more updates from each `run`
function. In go, the `crossdock.T` object is both, for reading and writing.

```
func myBehavior(t crossdock.T) {
  encoding := t.Param("encoding")
  // ..
  if something {
    t.Error("something failed")
  }
  t.Success("something else worked")
}
```

For Java, we can let the CrossdockRequest hold the params and write responses
to a CrossdockResponse object.

```
class CrossdockResponse {
  @Getter
  private List<TestResult> results;
  
  // The following append a TestResult into the results list with the correct
  // message and status.
  void success(String message);
  void error(String message);
  void skipped(String message);
  
  public static final class TestResult {
    private final String message;
    private final String status;
  }
}
```

Usage would be,

```
class MyBehavior {
  public CrossdockResponse run(CrossdockRequest req) {
    CrossdockResponse response = new CrossdockResponse();
    if (something) {
      response.error("something failed");
    }
    response.success("something else succeeded");
    return response;
  }
}
```

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/CrossdockParam.java
vivekrsharma: Yes, renamed to CrossdockRequest and implemented it the way we discussed.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/Behavior.java
vivekrsharma: Done.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
vivekrsharma: For now, need to skip the tests so I've added a TODO but agreed we could write an empty list.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockRequest.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package io.yarpc.crossdock;
+
+import java.util.List;
+import java.util.Map;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class CrossdockRequest {
+  private final Map<String, List<String>> params;
```
abhinav: Params sent by crossdock won't have multiple values. You should make this a `Map<String, String>`.

(Yeah, the HTTP spec allows for query parameters to hold a list of values but we don't do that in crossodock.)

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockResponse.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import lombok.Getter;
+
+public class CrossdockResponse {
+  @Getter private List<TestResult> results;
+
+  public CrossdockResponse() {
+    results = new ArrayList<>();
```
abhinav: Can delete constructor and do,

```
@Getter private final List<TestResult> results = new ArrayList();
```

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,56 @@
...
-@Slf4j
 public class CrossdockServerInboundHandler extends SimpleChannelInboundHandler<FullHttpRequest> {
-  private static final String SKIPPED = "skipped";
-  private static final String PASSED = "passed";
-  private static final String FAILED = "failed";
+  private final Map<String, Behavior> behaviors = new HashMap<String, Behavior>();
+
+  /** Populates all the different behaviors supported. */
+  public CrossdockServerInboundHandler() {
+    behaviors.put("raw", new EchoRawClient());
```
abhinav: Probably rename EchoRawClient to EchoRawBehavior

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,56 @@
...
-            + encoding);
 
-    // TODO: Make a client call and respond with appropriate message
+    Behavior br = behaviors.get(behavior);
+    CrossdockResponse response = new CrossdockResponse();
+    if (br != null) {
+      response = br.run(request);
+    } else {
+      // TODO: Throw an UnsupportedOperationException once json, thrift are fixed.
+      response.skipped("Thrift, Json not supported yet");
```
abhinav: Change to,

    response.skipped("Unsupported behavior: " + behavior)

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,56 @@
...
+    }
+    ObjectMapper objectMapper = new ObjectMapper();
+    String resBody = objectMapper.writeValueAsString(response.getResults().toArray());
```
abhinav: I'm not completely sure about this but I think you can just pass the `List<TestResult>` into `writeValueAsString` as-is without the `toArray`.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,56 @@
...
+    httpResponse =
+        new DefaultFullHttpResponse(
+            HttpVersion.HTTP_1_1,
+            HttpResponseStatus.OK,
+            Unpooled.wrappedBuffer(resBody.getBytes(Charsets.UTF_8)));
```
abhinav: If you're calling `getBytes` on the string anyway, you can use [`ObjectMapper.writeValueAsBytes`][1] instead.

[1]: https://fasterxml.github.io/jackson-databind/javadoc/2.5/com/fasterxml/jackson/databind/ObjectMapper.html#writeValueAsBytes(java.lang.Object)

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,56 @@
...
     QueryStringDecoder queryStringDecoder = new QueryStringDecoder(fullHttpRequest.uri());
-
-    // TODO: transport and encoding are not used, consider getting rid of them
-    String behavior = getQueryParam(queryStringDecoder, "behavior");
-    String server = getQueryParam(queryStringDecoder, "server");
-    String transport = getQueryParam(queryStringDecoder, "transport");
-    String encoding = getQueryParam(queryStringDecoder, "encoding");
+    CrossdockRequest request = populateRequest(queryStringDecoder);
+    String behavior = request.get("behavior");
     FullHttpResponse httpResponse = null;
```
abhinav: Nit: Drop this declaration. It doesn't need to be shared here and below.

```
if (behavior == null) {
 FullHttpResponse httpResponse = ...;
 channelHandlerContext.writeAndFlush(..);
 return;
}

// ...

String resBody = ...;
FullHttResponse httResponse = ...;
..
```

We don't need unrelated usages of the variables to share the value and risk
leaking the actual value between the two cases. Plus the code is easier to
follow if the variable you are using wasn't defined 50 lines above.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,56 @@
...
     channelHandlerContext.writeAndFlush(httpResponse);
     channelHandlerContext.pipeline().channel().close();
     channelHandlerContext.close();
   }
 
-  private String getQueryParam(QueryStringDecoder queryStringDecoder, String paramName) {
-    List<String> values = queryStringDecoder.parameters().get(paramName);
-    return values == null || values.isEmpty() ? null : values.get(0);
+  private CrossdockRequest populateRequest(QueryStringDecoder queryStringDecoder) {
+    Map<String, List<String>> params = queryStringDecoder.parameters();
```
abhinav: You'll want to squash the query params into a single `Map<String, String>`. Simply stomp over old values with newer values. That's exactly what [crossdock-go][1] does.

[1]: https://github.com/crossdock/crossdock-go/blob/master/client.go#L83

yarpc-crossdock/src/main/java/io/yarpc/crossdock/TestResult.java
```diff
@@ -21,11 +21,14 @@
...
-import lombok.NonNull;
+import lombok.Getter;
 
-@Data
-public class CrossdockTestResponse {
-  @NonNull private String status;
-  @NonNull private String output;
+public final class TestResult {
+  @Getter private String output;
+  @Getter private String status;
```
abhinav: final?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/TestResult.java
```diff
@@ -21,11 +21,14 @@
...
+
+  public TestResult(String output, String status) {
```
abhinav: `@RequiredArgsConstructor`?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
```diff
@@ -28,6 +28,10 @@
 
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public class CrossdockUtils {
+  public static final String SKIPPED = "skipped";
+  public static final String PASSED = "passed";
+  public static final String FAILED = "failed";
```
abhinav: Should `Status` be an enum?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockRequest.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public String get(String key) {
```
abhinav: getParam?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/raw/EchoRawClient.java
```diff
@@ -0,0 +1,64 @@
...
+  public CrossdockResponse run(CrossdockRequest request) throws Exception {
+    Dispatcher clientDispatcher = DispatcherBuilder.createDispatcher(request.get("server"));
+    CrossdockResponse crossdockResponse = new CrossdockResponse();
+    try {
+      clientDispatcher.start();
+      RawClient client = new RawClient(clientDispatcher.getClientConfig(SERVICE));
+      Reply<byte[]> reply =
+          client
+              .unary(ECHO_PROCEDURE, "abc".getBytes(Charsets.UTF_8))
+              .call(Context.builder().deadLine(Instant.now().plusSeconds(1)).build())
```
abhinav: Not related to this PR: This doesn't seem right. Clients should just specify timeouts, not deadlines. Deadlines should only be tracked by the library. At least an API should be exposed for setting a timeout which automatically infers the deadline. Also, it should be `deadline` not `deadLine`. CC @bpon

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
vivekrsharma: Tried using transforms on value - LMK if that looks okay to you.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/TestResult.java
vivekrsharma: Well, we're using this json serializer on this value and they can't be final if we want to use that.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
vivekrsharma: Do we prefer it to be strings? @bpon

abhinav: Yes, for the behaviors API we should use an enum because it enumerates all possible values for a status. We still need to translate it to a string before it gets serialized to JSON though.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockResponse.java
```diff
@@ -0,0 +1,42 @@
...
+ * THE SOFTWARE.
+ */
+package io.yarpc.crossdock;
+
+import java.util.ArrayList;
+import java.util.List;
+import lombok.Getter;
+
+public class CrossdockResponse {
+  @Getter private List<TestResult> results = new ArrayList<>();
```
bpon: could be final

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +31,57 @@
...
     channelHandlerContext.close();
   }
 
-  private String getQueryParam(QueryStringDecoder queryStringDecoder, String paramName) {
-    List<String> values = queryStringDecoder.parameters().get(paramName);
-    return values == null || values.isEmpty() ? null : values.get(0);
+  private CrossdockRequest populateRequest(QueryStringDecoder queryStringDecoder) {
+    Map<String, List<String>> queryParams = queryStringDecoder.parameters();
+    Map<String, String> params = new HashMap<>();
+    queryParams.forEach((key, value) -> params.put(key, (value != null ? value.get(0) : null)));
```
bpon: What about doing `params.put` only if `value` is not null or empty?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/raw/EchoRawBehavior.java
```diff
@@ -0,0 +1,64 @@
...
+    CrossdockResponse crossdockResponse = new CrossdockResponse();
+    try {
+      clientDispatcher.start();
+      RawClient client = new RawClient(clientDispatcher.getClientConfig(SERVICE));
+      Reply<byte[]> reply =
+          client
+              .unary(ECHO_PROCEDURE, "abc".getBytes(Charsets.UTF_8))
+              .call(Context.builder().deadLine(Instant.now().plusSeconds(1)).build())
+              .toCompletableFuture()
+              .get(1, TimeUnit.SECONDS);
```
bpon: We should not block execution inside a Netty channel.  I think this needs to return a `CompletionStage<CrossdockResponse>`.

docker-compose.yml
```diff
@@ -24,7 +25,7 @@ services:
             - AXIS_APACHETHRIFTCLIENT=go
             - AXIS_APACHETHRIFTSERVER=java
             - AXIS_GAUNTLET=go
-            - AXIS_HTTPSERVER=java,go
+            - AXIS_HTTPSERVER=go,java
```
abhinav: redundant change

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,87 @@
...
-@Slf4j
 public class CrossdockServerInboundHandler extends SimpleChannelInboundHandler<FullHttpRequest> {
-  private static final String SKIPPED = "skipped";
-  private static final String PASSED = "passed";
-  private static final String FAILED = "failed";
+  private final Map<String, Behavior> behaviors = new HashMap<String, Behavior>();
+
+  /** Populates all the different behaviors supported. */
+  public CrossdockServerInboundHandler() {
+    behaviors.put("raw", new EchoRawBehavior());
```
abhinav: This is fine for now but in the future we'll probably want the caller to register behaviors explicitly.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,87 @@
...
+                } else {
+                  CrossdockResponse crossdockResponse = new CrossdockResponse();
+                  crossdockResponse.error(ex.getMessage());
+                  writeAndCloseChannel(channelHandlerContext, crossdockResponse);
+                }
+                return true;
+              });
+    } else {
+      // TODO: Throw an UnsupportedOperationException once json, thrift are fixed.
+      CrossdockResponse response = new CrossdockResponse();
```
abhinav: Readability nit: Handle the failure case first and return early. This will allow the success case to be less nested.

```java
if (br == null) {
  response.skipped(...);
  writeAndCloseChannel(...);
  return;
}

br.run(request)
  .handle(..);
```

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,87 @@
...
+  private final Map<String, Behavior> behaviors = new HashMap<String, Behavior>();
+
+  /** Populates all the different behaviors supported. */
+  public CrossdockServerInboundHandler() {
+    behaviors.put("raw", new EchoRawBehavior());
+  }
 
   @Override
   protected void channelRead0(
       ChannelHandlerContext channelHandlerContext, FullHttpRequest fullHttpRequest)
```
abhinav: Nit: Rename `channelHandlerContext` to `ctx`. That's what we use everywhere else in the Netty code.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,87 @@
...
+                if (ex == null) {
+                  writeAndCloseChannel(channelHandlerContext, response);
+                } else {
+                  CrossdockResponse crossdockResponse = new CrossdockResponse();
+                  crossdockResponse.error(ex.getMessage());
+                  writeAndCloseChannel(channelHandlerContext, crossdockResponse);
+                }
+                return true;
+              });
+    } else {
```
abhinav: Nit: Looks like this whole block is an asynchronous operation that takes a `CrossdockRequest` and returns a `CrossdockResponse` which is eventually written to the channel. Consider changing it to something like,

```java
Behavior br = behaviors.get(behavior);
runBehavior(br, request)
  .handle((response, ex) -> {
    if (ex != null) {
      ctx.fireExceptionCaught(ex);
    } else {
       writeAndCloseChannel(ctx, response));
    }
  })
```

Where,

```java
private CompletionStage<CrossdockResponse> runBehavior(Behavior br, CrossdockRequest req) {
  CompletableFuture<CrossdockResponse> resFuture = new CompletableFuture<>();
  if (br == null) {
    CrossdockResponse res = new CrossdockResponse();
    res.skipped(...);
    resFuture.complete(res);
    return resFuture;
  }
  
  br.run(req)
    .handle((res, ex) -> {
      if (ex != null) {
        CrossdockResponse res = new CrossdockResponse();
        res.error(ex.getMessage());
      }
      resFuture.complete(res);
    })
  return resFuture;
}
```

What we avoid here is calls to `writeAndCloseChannel` in multiple places. Instead, it becomes "do this asynchronous thing and write its result to the channel".

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,87 @@
...
+      // TODO: Throw an UnsupportedOperationException once json, thrift are fixed.
+      CrossdockResponse response = new CrossdockResponse();
+      response.skipped("Unsupported behavior: " + behavior);
+      writeAndCloseChannel(channelHandlerContext, response);
+    }
+  }
+
+  private void writeAndCloseChannel(
+      ChannelHandlerContext channelHandlerContext, CrossdockResponse response) {
```
abhinav: "ctx"

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,87 @@
...
+      channelHandlerContext.writeAndFlush(httpResponse);
+    } catch (JsonProcessingException e) {
+      FullHttpResponse httpResponse =
+          new DefaultFullHttpResponse(
+              HttpVersion.HTTP_1_1,
+              HttpResponseStatus.OK,
+              Unpooled.wrappedBuffer(e.getMessage().getBytes(Charsets.UTF_8)));
+      channelHandlerContext.writeAndFlush(httpResponse);
+    }
     channelHandlerContext.pipeline().channel().close();
```
abhinav: IIRC, the `close` should be done through a listener on the `writeAndFlush` call.

    ctx.writeAndFlush(response).addListener(ChannelFutureListener.CLOSE);

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +32,87 @@
...
-  private String getQueryParam(QueryStringDecoder queryStringDecoder, String paramName) {
-    List<String> values = queryStringDecoder.parameters().get(paramName);
-    return values == null || values.isEmpty() ? null : values.get(0);
+  private CrossdockRequest populateRequest(QueryStringDecoder queryStringDecoder) {
+    Map<String, List<String>> queryParams = queryStringDecoder.parameters();
+    Map<String, String> params = new HashMap<>();
+    queryParams.forEach(
+        (key, value) -> {
+          if (value != null) {
+            params.put(key, (value != null ? value.get(0) : null));
```
abhinav: You have a `value != null` check above. This ternary operator is unnecessary.

abhinav: Also, change the check to,

    value != null && vaule.size() > 0

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/raw/EchoRawBehavior.java
```diff
@@ -0,0 +1,65 @@
...
+    RawClient client = new RawClient(clientDispatcher.getClientConfig(SERVICE));
+    return client
+        .unary(ECHO_PROCEDURE, "abc".getBytes(Charsets.UTF_8))
+        .call(Context.builder().deadLine(Instant.now().plusSeconds(1)).build())
+        .handle(
+            (result, ex) -> {
+              CrossdockResponse crossdockResponse = new CrossdockResponse();
+              if (ex != null) {
+                crossdockResponse.error(ex.getMessage());
+              } else {
```
abhinav: The thing responsible for running the behavior handles the exceptions. You can
handle only the success case here with `thenApply`.

```java
return client
  .unary(..)
  .call(...)
  .thenApply(result -> {
    String message = new String(result.getBody(), ...);
    // ...
    return crossdockResponse;
  })
```

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/raw/EchoRawBehavior.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final String ECHO_PROCEDURE = "echo/raw";
+
+  @Override
+  public CompletionStage<CrossdockResponse> run(CrossdockRequest request) {
+    Dispatcher clientDispatcher = DispatcherBuilder.createDispatcher(request.getParam("server"));
+    clientDispatcher.start();
+    RawClient client = new RawClient(clientDispatcher.getClientConfig(SERVICE));
+    return client
+        .unary(ECHO_PROCEDURE, "abc".getBytes(Charsets.UTF_8))
+        .call(Context.builder().deadLine(Instant.now().plusSeconds(1)).build())
```
abhinav: Created https://github.com/yarpc/yarpc-java/issues/291 to track this

docker-compose.yml
vivekrsharma: I thought we always had it alphabetically order and rebase messed it for me.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
vivekrsharma: Completion Stage guru! Done.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +33,94 @@
...
 import io.netty.handler.codec.http.QueryStringDecoder;
+import io.yarpc.crossdock.client.Behavior;
+import io.yarpc.crossdock.client.raw.EchoRawBehavior;
+import java.util.HashMap;
 import java.util.List;
-import lombok.extern.slf4j.Slf4j;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import org.apache.commons.codec.Charsets;
```
abhinav: This should probably be [Guava's Charsets class][1].

[1]: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Charsets.html

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockResponse.java
```diff
@@ -19,36 +19,24 @@
...
-   */
-  public String testEcho(ClientConfig config) throws Exception {
-    RawClient client = new RawClient(config);
-    Reply<byte[]> reply =
-        client
-            .unary("echo/raw", "abc".getBytes(Charsets.UTF_8))
-            .call(Context.builder().build())
-            .toCompletableFuture()
-            .get(1, TimeUnit.SECONDS);
+  public void error(String message) {
```
abhinav: We should make these methods return the CrossdockResponse itself so that we can chain it.

```java
return new CrossdockResponse().success("foo").error("bar")
```

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +33,94 @@
...
+              if (ex != null) {
+                ctx.fireExceptionCaught(ex);
+              } else {
+                writeAndCloseChannel(ctx, response);
+              }
+            });
+  }
+
+  private CompletionStage<CrossdockResponse> runBehavior(Behavior br, CrossdockRequest req) {
+    CompletableFuture<CrossdockResponse> resFuture = new CompletableFuture<>();
```
abhinav: resFuture is used only inside the if-block so declare it in there

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +33,94 @@
...
+    if (br == null) {
+      CrossdockResponse response = new CrossdockResponse();
+      response.skipped("Unsupported behavior: " + req.getParam("behavior"));
+      resFuture.complete(response);
```
abhinav: CrossdockResponse suggestion would simplify this to,

    return CompletableFuture.completedFuture(
      new CrossdockResponse()
        .skipped("Unsupperted behavior: " + req.getParam("behavior"))
    );

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +33,94 @@
...
+      return resFuture;
+    }
+
+    return br.run(req)
+        .handle(
+            (response, ex) -> {
+              if (ex != null) {
+                response.error(ex.getMessage());
```
abhinav: Are you sure this works? `response` will most likely be null if the underlying function threw an exception.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +33,94 @@
...
+      FullHttpResponse httpResponse =
+          new DefaultFullHttpResponse(
+              HttpVersion.HTTP_1_1, HttpResponseStatus.OK, Unpooled.wrappedBuffer(array));
+      ctx.writeAndFlush(httpResponse).addListener(ChannelFutureListener.CLOSE);
+    } catch (JsonProcessingException e) {
+      FullHttpResponse httpResponse =
+          new DefaultFullHttpResponse(
+              HttpVersion.HTTP_1_1,
+              HttpResponseStatus.INTERNAL_SERVER_ERROR,
+              Unpooled.wrappedBuffer(e.getMessage().getBytes(Charsets.UTF_8)));
```
abhinav: DRY this up a bit

```
byte[] body;
HttpResponseStatus status = ...
try {
  body = ..
} catch (..) {
  status = HttpResponseStatus.INTERNAL_SERVER_ERROR;
  body = ..
}

FullHttpResponse httpResponse = new DefaultFullHttpResponse(...);
```

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/raw/EchoRawBehavior.java
```diff
@@ -0,0 +1,62 @@
...
+  private static final String ECHO_PROCEDURE = "echo/raw";
+
+  @Override
+  public CompletionStage<CrossdockResponse> run(CrossdockRequest request) {
+    Dispatcher clientDispatcher = DispatcherBuilder.createDispatcher(request.getParam("server"));
+    clientDispatcher.start();
+    RawClient client = new RawClient(clientDispatcher.getClientConfig(SERVICE));
+    return client
+        .unary(ECHO_PROCEDURE, "abc".getBytes(Charsets.UTF_8))
+        .call(Context.builder().deadline(Instant.now().plusSeconds(1)).build())
```
abhinav: `Context.timeout` is merged.  Rebase and use that instead:

    Context.builder().timeout(Duration.ofSeconds(1)).build()

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockServerInboundHandler.java
```diff
@@ -29,50 +34,87 @@
...
+  }
+
+  private CompletionStage<CrossdockResponse> runBehavior(Behavior br, CrossdockRequest req) {
+    if (br == null) {
+      return CompletableFuture.completedFuture(
+          new CrossdockResponse().skipped("Unsupported behavior: " + req.getParam("behavior")));
+    }
+
+    return br.run(req)
+        .handle(
```
bpon: You can use `exceptionally` instead of `handle` because you're not modifying the result in the success case.

PR #279, opened 2017-03-15
core: Unwrap CompletionExceptions

bpon:
When an unchecked exception is thrown in a completion stage handler, the exception is wrapped in a java.util.concurrent.CompletionException.


yarpc-core/src/main/java/io/yarpc/core/transport/exceptions/RequestEncodingException.java
```diff
@@ -73,6 +73,7 @@ public static RequestEncodingException newMismatch(
       TransportRequest request, String procedureEncoding) {
     return fromTransportRequest(
         request,
+        procedureEncoding,
```
abhinav: Good find. Although this change looks unrelated to this PR.

bpon: Oh yeah, sorry.  This should have gone into #280.  Juggling too many PRs!

PR #280, opened 2017-03-15
json/thrift: Throw transport-level exception when failing to decode request body

bpon:
Use the already defined RequestEncodingException.


yarpc-core/src/main/java/io/yarpc/core/transport/exceptions/RequestEncodingException.java
```diff
@@ -25,21 +25,11 @@
 
 /** Error thrown when trying to decode a request body. */
 public class RequestEncodingException extends BadRequestException {
-  public RequestEncodingException(
+  private RequestEncodingException(
```
abhinav: should this be RequestDecodingException?

yarpc-encoding-json/src/main/java/io/yarpc/encoding/json/UnaryJsonTransportHandler.java
```diff
@@ -60,4 +62,12 @@
                             objectMapper.writeValueAsBytes(reply.getBody())))
                     .build());
   }
+
+  private I readJson(TransportRequest req, byte[] input) {
+    try {
+      return objectMapper.readValue(input, klass);
+    } catch (IOException e) {
+      throw RequestEncodingException.fromTransportRequest(req, e.getMessage());
```
abhinav: Looks like we're losing some information here by discarding e. Should we
connect that up in case the RequestEncodingException gets logged before being
converted into a transport level error?

yarpc-core/src/main/java/io/yarpc/core/transport/exceptions/RequestEncodingException.java
bpon: Sure.  I was thinking of "encoding" as a noun, not a verb, but I see how that could be confusing.

PR #281, opened 2017-03-15
json/thrift: throw ResponseEncodingException for non-serializable responses

bpon:


yarpc-encoding-thrift/src/main/java/io/yarpc/encoding/thrift/UnaryThriftTransportHandler.java
```diff
@@ -80,4 +79,13 @@ private I readThrift(TransportRequest req, byte[] input) {
       throw RequestEncodingException.fromTransportRequest(req, e.getMessage());
     }
   }
+
+  private byte[] writeThrift(TransportRequest req, O obj) {
+    TSerializer serializer = new TSerializer(protocolFactory);
+    try {
+      return serializer.serialize(obj);
+    } catch (TException e) {
+      throw ResponseEncodingException.fromTransportRequest(req, e.getMessage());
```
abhinav: Same as the other PR. Should we still maintain `e` as the cause in the ResponseEncodingException so that if this is logged, we know the root cause.

bpon: ok

PR #282, opened 2017-03-17
http: Support multiple inbound requests over the same connection

bpon:
When an inbound request is received, check if it has a `Connection: close` header.  If so, then close the connection after sending a response.  Otherwise, leave the connection open.

Note:  We may need to explicitly set a connection timeout so that the connection doesn't stay open indefinitely.  I'm not sure what Netty's default is.


yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -185,7 +184,7 @@ protected void channelRead0(ChannelHandlerContext ctx, TransportRequest request)
               DefaultTransportResponse.builder()
                   .body(TransportBody.fromByteArray(new byte[0]))
                   .build();
-          ctx.writeAndFlush(response).addListener(ChannelFutureListener.CLOSE);
+          ctx.writeAndFlush(response).addListener(ChannelFutureListeners.CLOSE_IF_REQUESTED);
```
vivekrsharma: When does this get closed? Do we want to add a default timeout or something similar? I'm fine following that in a separate PR.

bpon: Yes, I'll look into it and submit another PR.

yarpc-integration-tests/src/test/java/io/yarpc/transport/http/HttpConnectionReuseTest.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  @After
+  public void releaseConnection() throws Exception {
+    conn.close();
+    connManager.releaseConnection(conn, null, 0, TimeUnit.SECONDS);
+  }
+
+  @Test
+  public void testConnectionReuse() throws Exception {
```
abhinav: nice

PR #283, opened 2017-03-17
crossdock: Enable errors behavior

bpon:
Enabling the errors behavior allowed me to discover a bunch of bugs related to exception handling, so I also fixed those and added/updated integration tests.


yarpc-core/src/main/java/io/yarpc/core/transport/DefaultTransportRequest.java
```diff
@@ -29,13 +29,13 @@
 @Getter
 @Builder
 public class DefaultTransportRequest implements TransportRequest {
-  @NonNull private String caller;
```
bpon: This is validated by the builder.  Having `@NonNull` here prevents us from even building the object, so we wouldn't be able to generate a meaningful "missing params" error.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpStatusMapper.java
```diff
@@ -56,7 +56,7 @@ public static TransportErrorResponse toTransportErrorResponse(
         body.accumulate()
             .map(
                 bytes -> {
-                  String message = new String(bytes, Charsets.UTF_8);
+                  String message = new String(bytes, Charsets.UTF_8).trim();
```
bpon: This remove the newline added during transport.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
```diff
@@ -210,7 +210,7 @@ public void onSuccess(TransportResponse response) {
 
     @Override
     public void onError(Throwable cause) {
-      ctx.fireExceptionCaught(cause);
+      new ErrorResponseEncoder().writeErrorResponse(ctx, cause);
```
bpon: All the channel handlers may have already executed at this point, so an exceptionCaught event might not get caught.  We can simply write the http response here.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/common/PhoneHandler.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public CompletionStage<Reply<PhoneResponse>> handleUnary(Context ctx, PhoneRequest phoneRequest) {
+    String encoding =
+        dispatcher
+            .getProcedures()
+            .stream()
+            .filter(procedure -> matchesPhoneRequest(procedure, phoneRequest))
+            .map(Procedure::getEncoding)
+            .findAny()
+            .orElse(JsonEncoding.ENCODING);
```
abhinav: Just use JSON. I don't think the phone endpoint is called with anything else.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpInbound.java
abhinav: should we inject the `ErrorResponseEncoder` we create above into this class? If we add more attributes/configuration on it, we'll probably want to re-use the same ErrorResponseEncoder.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/common/PhoneHandler.java
bpon: It is called with thrift sometimes.

abhinav: Are you referring to the call from the error behavior? That's a test for encoding mismatch. I don't think there's any valid call to `phone` where the procedure being called is not JSON.

Worth noting that the Go implementation always uses a JSON client. https://github.com/yarpc/yarpc-go/blob/dev/internal/crossdock/server/yarpc/phone.go#L98

bpon: Not sure how that works in Go, but...

```go
		{
			name: "remote bad request",
			headers: map[string]string{
				"RPC-Caller":     "yarpc-test",
				"RPC-Service":    "yarpc-test",
				"RPC-Procedure":  "phone",
				"RPC-Encoding":   "json",
				"Context-TTL-MS": "100",
			},
			body: `{
				"service": "yarpc-test",
				"procedure": "Echo::echo",
				"body": "not a Thrift payload",
				"transport": {"http": {"host": "` + t.Param(params.Server) + `", "port": 8081}}
			}`,
			wantStatus: 500,
			wantBodyStartsWith: `UnexpectedError: error for procedure "phone" of service "yarpc-test": ` +
				`BadRequest: failed to decode "thrift" request body for procedure "Echo::echo" ` +
				`of service "yarpc-test" from caller "yarpc-test": `,
		},
```

abhinav: Yeah, that's a test for a bad request. The Phone implementation will end up calling the Thrift endpoint `Echo::echo` with the body `not a Thrift payload`. It'll end up using `RPC-Encoding: json` because a `json.Client` is being used.

bpon: We validate headers before we decode the request body, so I think we would get an encoding mismatch if we tried to use a json client.

bpon: Ah you're right, it does work with json encoding.  I guess I over-complicated it.

PR #288, opened 2017-03-20
crossdock: enable json oneway behavior

vivekrsharma:


yarpc-crossdock/build.gradle
```diff
@@ -13,6 +13,8 @@ dependencies {
     
     compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.2'
 
+    compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.1.5'
```
abhinav: wut? two entries?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
```diff
@@ -21,18 +21,59 @@
...
   public static String getLocalhostUrl(int port) throws UnknownHostException {
     return String.format("http://%s:%s", InetAddress.getLocalHost().getHostAddress(), port);
   }
 
   public static String generateHostUrl(String host, int port) {
     return String.format("http://%s:%s", host, port);
   }
+
+  /**
+   * Makes a one way call using a RAW encoding.
```
abhinav: "raw" or "Raw" should be fine

abhinav: also, "oneway", not "one way"

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
```diff
@@ -21,18 +21,59 @@
...
+   *
+   * @param ctx context to use to make the call
+   * @param body body to pass
+   * @return ack to confirm the sent message
+   */
+  public static Single<Ack> callHome(Context ctx, byte[] body) {
```
abhinav: Explain the things it relies on in the Javadoc. Especially the part where it relies on a specific header to know which outbound to call. Also mention thigns like which procedure is called and what the caller and service names are.

abhinav: Also, rename this to `httpOnewayCallback` or similar to indicate that this is
actually making a call back to the original caller.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/oneway/JsonOnewayHandler.java
```diff
@@ -0,0 +1,47 @@
...
+/*
+ * Copyright (c) 2017, Uber Technologies, Inc
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
```
abhinav: This goes above the package declaration

yarpc-crossdock/build.gradle
vivekrsharma: sorry that was rebase trouble.Fixed.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
```diff
@@ -21,18 +21,63 @@
...
+   * It calls call-back procedure in oneway-client service.
+   * It uses raw encoding and callee is oneway-server.
+   * The outbound is derived from callBack headers
+   * passed in the context.
+   *
+   * @param ctx context to use to make the call
+   * @param body body to pass
+   * @return ack to confirm the sent message
+   */
+  public static Single<Ack> httpOnewayCallback(Context ctx, byte[] body) {
```
abhinav: Since this is used exclusively for the oneway tests, let's move it to a new class. We can call it `OnewayBehaviorUtils` or similar. And then you can rename the method to `httpCallback`.

abhinav: Also, since the result is always adapted to a CompletionStage, make this return that too.

PR #289, opened 2017-03-21
crossdock: enable timeout behavior

bpon:


yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/raw/SleepRawHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import io.yarpc.core.encoding.UnaryRequestHandler;
+import java.util.concurrent.CompletionStage;
+
+public class SleepRawHandler implements UnaryRequestHandler<byte[], byte[]> {
+  @Override
+  public CompletionStage<Reply<byte[]>> handleUnary(Context ctx, byte[] data) {
+    try {
+      Thread.sleep(1000);
+      return null;
+    } catch (InterruptedException e) {
```
abhinav: Instead of blocking the IO loop, should we instead return a CompletionStage
that resolves after a second?

Something like,

```
CompletableFuture<Reply<byte[]>> future = new CompletableFuture<>();
doAfter(Duration.seconds(1), () -> future.complete(new byte[] {}));
return future;
```

I'm not entirely sure what the best way to implement `doAfter` is. Probably
one of,

-   On-demand thread
-   Injected executor service
-   The Netty Channel's event loop (http://stackoverflow.com/a/16766323/58536)

bpon: I think an on-demand thread is good enough.  Don't have to get too fancy here.

PR #293, opened 2017-03-21
core: make Context immutable

bpon:


yarpc-core/src/main/java/io/yarpc/core/Context.java
```diff
@@ -27,13 +27,14 @@
...
+import lombok.Value;
 
-@Data
+@Value
 @Builder
 public class Context {
-  private Instant deadline;
-  private Map<String, String> headers;
+  private final Instant deadline;
+  @NonNull private final Map<String, String> headers;
```
abhinav: do we want headers to be mutable after the fact?

if not, we should maybe override the getter for headers to return a `Collections.unmodifiableMap` view around it.

yarpc-encoding-json/src/test/java/io/yarpc/encoding/json/JsonClientTest.java
```diff
@@ -70,7 +69,7 @@ public void testCallUnary() throws Exception {
     Data res =
         client
             .unary("procedure", new Data(123), Data.class)
-            .call(ctx)
+            .call(Context.builder().build())
```
abhinav: Thoughts on adding an `empty` static constructor for this case? `Context.empty()` reads a lot nicer than `Context.builder().build()`.

abhinav: Feel free to ignore this if this will be moot once we start connecting parent-child contexts

bpon: Sure, I'll do that.

PR #299, opened 2017-03-23
json: allow null body

vivekrsharma:
Testing with python exposed that body could be null.
Allowing a null body and verifying with a test case.


vivekrsharma:
@willhug Thought you might be interested in this :)

yarpc-integration-tests/src/test/java/io/yarpc/encoding/json/JsonIntegrationTest.java
```diff
@@ -116,6 +116,20 @@ public void testUnaryCall() throws Exception {
...
+    Reply<KeyValue> reply =
+        client
+            .unary(ECHO_PROCEDURE, null, KeyValue.class)
+            .header("hello", "world")
+            .call(Context.empty())
+            .toCompletableFuture()
+            .get();
+
+    assertEquals("world", reply.getHeaders().get("hello"));
+    assertEquals(null, reply.getBody());
```
bpon: You could use `assertNull` if you want.

vivekrsharma: Sure. I'm so excited. The # of tests is like 450 now!

PR #303, opened 2017-03-24
http: Add tracing

bpon:
Using the [opentracing](http://opentracing.io) API


yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportRequestDecoderConfiguration.java
```diff
@@ -66,25 +79,41 @@ public TransportRequest build(
...
-      return now.plusMillis(Long.parseLong(timeout));
+      return start.plusMillis(Long.parseLong(timeout));
     } catch (NumberFormatException e) {
       throw new InvalidTimeoutException(service, procedure, timeout, e);
     }
   }
 
+  private Span createSpan(
+      Instant start, String service, String procedure, String caller, HttpHeaders httpHeaders) {
+    SpanBuilder spanBuilder =
```
peter-edge: nit: wout prefer tracer be in the same line

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportRequestDecoderConfiguration.java
```diff
@@ -66,25 +79,41 @@ public TransportRequest build(
...
+        tracer
+            .buildSpan(procedure)
+            .withStartTimestamp(Instant.EPOCH.until(start, ChronoUnit.MICROS))
+            .withTag(SpanTags.CALLER, caller)
+            .withTag(SpanTags.SERVICE, service)
+            .withTag(SpanTags.ENCODING, httpHeaders.get(HeaderMapper.ENCODING))
+            .withTag(SpanTags.TRANSPORT, "http");
```
peter-edge: nit: http as private static final?

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportRequestDecoderConfiguration.java
```diff
@@ -66,25 +79,41 @@ public TransportRequest build(
...
+    SpanContext parent =
```
peter-edge: same line, newlines after each parameter?

yarpc-transport-http/src/test/java/io/yarpc/transport/http/TransportRequestDecoderTest.java
```diff
@@ -69,7 +71,8 @@ public void before() {
               resultTester = req.getBody().accumulate().test();
               return invocation.getMock();
             });
-    decoder = new HttpTransportDecoder<>(TransportRequestDecoder.CONFIGURATION);
+    Tracer tracer = NoopTracerFactory.create();
+    decoder = new HttpTransportDecoder<>(new TransportRequestDecoderConfiguration(tracer));
```
peter-edge: nit: just put `NoopTracerFactory.create()` in the `TransportRequestDecoderConfiguration` constructor, this is similar to other parts of the code here

yarpc-transport-http/src/test/java/io/yarpc/transport/http/TransportRequestEncoderTest.java
```diff
@@ -56,14 +59,16 @@
...
   @Before
   public void before() throws Exception {
-    encoder =
-        new HttpTransportEncoder<>(
-            new TransportRequestEncoderConfiguration(
-                new URL("http://test.uber.internal:12345/foo")));
+    Tracer tracer = NoopTracerFactory.create();
+    encoderConfig =
+        new TransportRequestEncoderConfiguration(
+            new URL("http://test.uber.internal:12345/foo"), tracer);
```
peter-edge: nit: same

yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpTransport.java
```diff
@@ -48,6 +51,8 @@
   @NonNull @Getter private final NioEventLoopGroup workerGroup = new NioEventLoopGroup();
   @NonNull @Getter private final NioEventLoopGroup bossGroup = new NioEventLoopGroup();
 
+  @NonNull @Getter private final Tracer tracer;
```
bpon: Sure.  The only downside is that it introduces a dependency on opentracing-noop, but it has no external dependencies, so not a big deal.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportRequestDecoderConfiguration.java
bpon: I would like to, but the auto-formatter does this.

yarpc-transport-http/src/main/java/io/yarpc/transport/http/TransportRequestDecoderConfiguration.java
bpon: I don't think that making this a private static final improves readability in this case.

I would really like for this to be `HttpTransport.NAME`, but for some reason we use lowercase here and uppercase everywhere else (which is consistent with yarpc-go).  We could also do `HttpTransport.NAME.toLowerCase()` but the `toLowerCase()` part seems kind of arbitrary, so I'd prefer to leave this as is.

PR #304, opened 2017-03-27
crossdock: enable header behavior

vivekrsharma:


yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
```diff
@@ -0,0 +1,117 @@
...
+        HeaderTestCase.data()
+            .stream()
+            .map(
+                h ->
+                    caller
+                        .call(h.getGive())
+                        .thenApply(
+                            i ->
+                                Collections.singletonList(
+                                    new ResultItem(i, h.getWant(), h.getDesc()))))
```
kriskowal: Don’t know whether you’re blazing new Java style territory, but it would be easier to read this in a more compact and less lispy dialect:

```
.data()
.stream()
.map(h ->
   caller
   .call(h.getGive())
   .thenApply(i ->
       Collections.singletonList(
           new ResultItem(i, h.getWant(), h.getDesc())
       )
    )
)
.collect(Collectors.toList());
```

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
```diff
@@ -0,0 +1,117 @@
...
+            .collect(Collectors.toList());
+
+    final CompletableFuture<List<ResultItem>>[] all = new CompletableFuture[stages.size()];
+    for (int i = 0; i < stages.size(); i++) {
+      all[i] = stages.get(i).toCompletableFuture();
+    }
+    return CompletableFuture.allOf(all)
+        .thenApply(
+            ignored -> {
```
kriskowal: I expect that the `ignored` here should be a collection of the completion values of `all` completable futures. It shouldn’t be necessary to join each of the `all` elements manually like this. However, I did see an example that streamed and collected `all`. If `allOf` doesn’t do the job, we should roll one that does.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
```diff
@@ -0,0 +1,117 @@
...
+                }
+              }
+              return crossdockResponse;
+            });
+  }
+
+  private boolean compareHeaders(Map<String, String> want, Map<String, String> got) {
+    for (Map.Entry<String, String> entry : want.entrySet()) {
+      if (!got.containsKey(entry.getKey())
+          || !want.get(entry.getKey()).equalsIgnoreCase(entry.getValue())) {
```
kriskowal: I expect operators to hang at the end of the prior line, generally. Not just commas. Let me know if that’s outside our Java style guide.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
bpon: The lambdas can be extracted into methods to make this block more compact and readable.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
```diff
@@ -0,0 +1,117 @@
...
+        break;
+        // TODO(viveksharma): add json and thrift support here.
+      default:
+        return CompletableFuture.completedFuture(
+            new CrossdockResponse()
+                .skipped(
+                    "Unsupported headers behavior for encoding:" + request.getParam("encoding")));
+    }
+
+    List<CompletionStage<List<ResultItem>>> stages =
```
bpon: This can be a `List<CompletionStage<ResultItem>>`.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
```diff
@@ -0,0 +1,117 @@
...
+                h ->
+                    caller
+                        .call(h.getGive())
+                        .thenApply(
+                            i ->
+                                Collections.singletonList(
+                                    new ResultItem(i, h.getWant(), h.getDesc()))))
+            .collect(Collectors.toList());
+
+    final CompletableFuture<List<ResultItem>>[] all = new CompletableFuture[stages.size()];
```
bpon: You can just call `toArray()` on the `Stream` above.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
```diff
@@ -0,0 +1,117 @@
...
+              final List<ResultItem> result = new ArrayList<>(all.length);
+              for (int i = 0; i < all.length; i++) {
+                result.addAll(all[i].join());
+              }
+              return result;
+            })
+        .handle(
+            (list, ex) -> {
+              CrossdockResponse crossdockResponse = new CrossdockResponse();
+              for (ResultItem item : list) {
```
bpon: `list` could be null if there is an exception.

bpon: I think you might want to just roll this whole block up into the `thenApply` block.  I'm not sure why you're using `handle` when you're not handling exceptions.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
bpon: `want.get(entry.getKey())` is the same as `entry.getValue()`.  I think you mean `got.get(...)`

bpon: nit:  Can you rename `entry` to something else?  There are 2 maps here, and I had to read this like 3 times because I kept forgetting which one you're referring to.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderTestCase.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.Map;
+import java.util.UUID;
+import lombok.Builder;
+import lombok.Data;
+
+@Builder
+@Data
+public class HeaderTestCase {
+
+  public static final String TOKEN1 = UUID.randomUUID().toString();
```
bpon: Shouldn't these be re-generated for each test?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/ResultItem.java
```diff
@@ -0,0 +1,34 @@
...
+ * THE SOFTWARE.
+ */
+package io.yarpc.crossdock.client.headers;
+
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@AllArgsConstructor
+@Data
```
bpon: I think you mean `@Value`?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderTestCase.java
vivekrsharma: nope, even in Go we just generate it once and stick same token in all the tests.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
vivekrsharma: Yes, this is the art of dealing with lambdas. In general, lambdas are pretty unreadable and chaining like this can make it worse.  Is there a rule we apply or is it whatever feels right?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/HeaderBehavior.java
vivekrsharma: Yes, I tried couple of ways with @bpon but got only this work. Also note that the join is called from inside of thenApply that implies it wouldn't be blocking.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/ResultItem.java
```diff
@@ -0,0 +1,34 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.yarpc.crossdock.client.headers;
+
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import lombok.Value;
+
+@AllArgsConstructor
```
bpon: `@Value` already includes `@AllArgsConstructor`

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/RawCaller.java
```diff
@@ -0,0 +1,49 @@
...
+import io.yarpc.encoding.raw.RawClient;
+import java.time.Duration;
+import java.util.Map;
+import java.util.concurrent.CompletionStage;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public class RawCaller implements HeaderCaller {
+  private static String ECHO_PROCEDURE = "echo/raw";
+  private RawClient client;
```
bpon: nit: final (and `@RequiredArgsConstructor`)

PR #305, opened 2017-03-27
crossdock: enable json header behavior

vivekrsharma:


yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/JsonCaller.java
```diff
@@ -0,0 +1,52 @@
...
+import io.yarpc.encoding.json.JsonClient;
+import java.time.Duration;
+import java.util.Map;
+import java.util.concurrent.CompletionStage;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public class JsonCaller implements HeaderCaller {
+  private static String JSON_PROCEDURE = "echo";
+  private JsonClient client;
```
bpon: nit: final

Then you could use `@RequiredArgsConstructor` instead of `@AllArgsConstructor`, but not a big deal.

PR #306, opened 2017-03-27
crossdock: enable thrift header behavior

vivekrsharma:


yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/headers/ThriftCaller.java
```diff
@@ -0,0 +1,48 @@
...
+import io.yarpc.thrift.generated.yarpc.Echo;
+import io.yarpc.thrift.generated.yarpc.Echo.EchoReply;
+import java.time.Duration;
+import java.util.Map;
+import java.util.concurrent.CompletionStage;
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public class ThriftCaller implements HeaderCaller {
+  private Echo.Client client;
```
bpon: nit: final

Then you could use `@RequiredArgsConstructor` instead of `@AllArgsConstructor`, but not a big deal.

PR #307, opened 2017-03-27
thrift-generator: Fix conversion of collection and exception fields

bpon:
Fixes a couple of issues:
- Collection types (set, list, map) were being converted to `java.util` collection types using their simple class name.  Corrected this to use the fully qualified class name.  Another option would have been to add imports for all Java collection types, but that carries a risk of class name collision.
- Exception fields were being named by the exception type instead of the field name/id.  Corrected this by looking up exceptions by Thrift field id.

These bugs were caught by including `gauntlet.thrift` in tests.


yarpc-thrift-generator/src/main/resources/templates/yarpc/common.stg
```diff
@@ -47,7 +47,8 @@ _clientMethod(method) ::= <<
 public io.yarpc.core.Call\<<if(method.oneway)>io.yarpc.core.transport.Ack<else><_reply(method)><endif>\> <method.javaName>(
     <if(method.parameters)><method.parameters: {param |<_param(param)>}; separator=",\n"><endif>
 ) {
-    <_args(method)> args = new <_args(method)>(<if(method.parameters)><method.parameters: {param |<param.javaName>}; separator=", "><endif>);
+    <_args(method)> args = new <_args(method)>()<if(method.parameters)>
```
bpon: Sure, let's add some comments later.

PR #309, opened 2017-03-28
http: Shutdown NioEventLoopGroups when stopping the transport

bpon:
We were previously stopping them when stopping Outbounds and Inbounds, which means Netty would be unable to make/serve additional requests after any Outbound or Inbound has been stopped.


yarpc-transport-http/src/main/java/io/yarpc/transport/http/HttpOutbound.java
```diff
@@ -107,7 +107,7 @@ private void doStart() {
 
   @Override
   public void stop() {
-    lifecycle.stop(workerGroup::shutdownGracefully);
+    lifecycle.stop(() -> {});
```
breerly: What am I looking at here, specifically `() -> {}` passed as a param.

bpon: This line reads: Set the lifecycle state to stopped and perform a no-op as the stop action.

breerly: I see, thats an anon func with an empty impl. 👍

PR #310, opened 2017-03-29
crossdock: Enable context propagation behavior

bpon:


yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
```diff
@@ -40,6 +47,49 @@ public static String generateHostUrl(String host, int port) {
...
+      }
+      ctx.baggageItems().forEach(item -> injectItem(item, (TextMap) carrier));
+    }
+
+    private void injectItem(Map.Entry<String, String> item, TextMap textMap) {
+      if (TRACE_ID.equalsIgnoreCase(item.getKey())) {
+        // Save trace id as a String because it might not be a number, as MockContext expects
+        textMap.put(TRACE_ID, item.getValue());
+      } else {
+        textMap.put(BAGGAGE_PREFIX + item.getKey(), item.getValue());
```
peter-edge: you could be super cool and do a wildcard here, but that's probably too cute:

```java
textMap.Put(TRACE_ID.equalsIgnoreCase(item.GetKey()) ? TRACE_ID : BAGGAGE_PREFIX + item.GetKey(), item.GetValue())
```

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
```diff
@@ -40,6 +47,49 @@ public static String generateHostUrl(String host, int port) {
...
+      }
+    }
+
+    @Override
+    public <C> MockContext extract(Format<C> format, C carrier) {
+      if (!(carrier instanceof TextMap)) {
```
peter-edge: can you explain why this `instanceof` logic is being used through this class?

bpon: Oops, that's totally unnecessary.  We can safely assume the carrier is a TextMap.

PR #311, opened 2017-03-30
crossdock: enable oneway behavior

vivekrsharma:
Currently, enables crossdock for oneway behavior and have support for raw encoding. PRs for json and thrift to follow.


yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
```diff
@@ -35,4 +36,8 @@ public static String getLocalhostUrl(int port) throws UnknownHostException {
   public static String generateHostUrl(String host, int port) {
     return String.format("http://%s:%s", host, port);
   }
+
+  public static int getRandomPort() {
+    return new Random().ints(33000, 60000).iterator().next();
```
peter-edge: talked offline, filing an issue to write a function to find a free port, not assuming we are just in crossdock and all ports are free in this range

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/dispatcher/DispatcherBuilder.java
```diff
@@ -45,8 +51,39 @@ public static Dispatcher createDispatcher(String serverHostName) {
...
+   */
+  public static Dispatcher createOnewayDispatcher(
+      String serverHostName, CompletableFuture<byte[]> future, int listenPort) {
+    HttpTransport httpTransport = new HttpTransport();
+    log.debug(
+        "The url wer're calling is [" + String.format("http://%s:8084", serverHostName) + "]");
+    HttpOutbound outbound =
+        httpTransport
+            .outboundBuilder()
+            .url(String.format("http://%s:8084", serverHostName))
```
peter-edge: is there any chance in the future we could get in trouble with https?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/oneway/RawCaller.java
```diff
@@ -0,0 +1,64 @@
...
+import io.yarpc.crossdock.CrossdockResponse;
+import io.yarpc.encoding.raw.RawClient;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import lombok.RequiredArgsConstructor;
+
+@RequiredArgsConstructor
+public class RawCaller implements OnewayCaller {
+  private static String ECHO_PROCEDURE = "echo/raw";
```
peter-edge: this might not be possible because of crossdock, but I'd like if this was changed to `echo::raw` across all systems later, to be consistent with thrift (and protobuf)

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
vivekrsharma: https://github.com/yarpc/yarpc-java/issues/312

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/dispatcher/DispatcherBuilder.java
vivekrsharma: i'm not sure. But this would need changing it all over. Can followup once we have https support?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/CrossdockUtils.java
bpon: Maybe do something like this to check that the port is free? https://github.com/yarpc/yarpc-java/blob/master/yarpc-integration-tests/src/test/java/io/yarpc/TestUtil.java#L45

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/dispatcher/DispatcherBuilder.java
```diff
@@ -45,8 +51,35 @@ public static Dispatcher createDispatcher(String serverHostName) {
...
+    HttpInbound inbound = httpTransport.inboundBuilder().port(listenPort).build();
+    Dispatcher clientDispatcher =
+        io.yarpc.core.Dispatcher.builder()
+            .name("oneway-client")
+            .onewayOutbound("oneway-server", outbound)
+            .inbound(inbound)
+            .build();
+    clientDispatcher.register(
+        RawEncoding.newOnewayProcedure("call-back", new OnewayRawHandler(future)));
+    clientDispatcher.start();
```
bpon: Seems a little weird for the dispatcher to be started here.  Maybe move this to `OnewayBehavior.run`?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/dispatcher/DispatcherBuilder.java
```diff
@@ -45,8 +51,35 @@ public static Dispatcher createDispatcher(String serverHostName) {
...
+            .build();
+
+    HttpInbound inbound = httpTransport.inboundBuilder().port(listenPort).build();
+    Dispatcher clientDispatcher =
+        io.yarpc.core.Dispatcher.builder()
+            .name("oneway-client")
+            .onewayOutbound("oneway-server", outbound)
+            .inbound(inbound)
+            .build();
+    clientDispatcher.register(
```
bpon: Doesn't this belong in the behavior class?  Different behaviors could potentially depend on different procedures.

vivekrsharma: nope. All all behavior use raw handling to get the call-back.

PR #313, opened 2017-03-30
crossdock: enable oneway json behavior

vivekrsharma:

PR #314, opened 2017-03-31
crossdock: enable oneway thrift behavior

vivekrsharma:


yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/oneway/OnewayBehavior.java
```diff
@@ -50,6 +51,9 @@
       case "json":
         caller = new JsonCaller(new JsonClient(clientDispatcher.getClientConfig(SERVICE)));
         break;
+      case "thrift":
+        caller = new ThriftCaller(new Client(clientDispatcher.getClientConfig(SERVICE)));
```
bpon: nit: `Client` is too generic in this context.  I like `Oneway.Client` better.

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/oneway/ThriftCaller.java
```diff
@@ -0,0 +1,46 @@
...
+public class ThriftCaller implements OnewayCaller {
+  private final Oneway.Client client;
+
+  @Override
+  public CompletionStage<CrossdockResponse> call(
+      CompletableFuture<byte[]> future, int servicePort) {
+    Call<Ack> call =
+        client.echo("abc").header("callBackAddr", String.format("java:%d", servicePort));
+    call.call(Context.builder().timeout(Duration.ofSeconds(1)).build());
+    return future.handle((result, ex) -> OnewayUtils.processResponse(result, ex));
```
bpon: nit: Same as in #313 : this could be `OnewayUtils::processResponse`

Issue #302, opened 2017-03-24
gradle: Improve dependency management across subprojects


bpon:
I have used the [dependency-management plugin](https://github.com/spring-gradle-plugins/dependency-management-plugin) for other Gradle projects.  It's useful for managing versions at the root project level.  I'm not aware of any other options.

PR #316, opened 2017-03-31
gradle: Manage dependency versions in root build.gradle

bpon:
Use the [dependency-management plugin](https://github.com/spring-gradle-plugins/dependency-management-plugin) to manage versions of commonly used dependencies across projects.


bpon:
Fixes #302

PR #321, opened 2017-04-03
crossdock: enable gauntlet

vivekrsharma:
Enables the thrift gauntlet behavior with java as server.
Currently the tests fail due to violation of Numberz being
an enum. All the other tests pass with expected behavior.


breerly:
This is amazing! Can you get the build passing, or should we review anyways?

bpon:
It looks like the crossdock failures are due to thriftrw's generated Go code and Apache Thrift's generated Java code handling enums differently.

Given this enum in thrift:
```thrift
enum Numberz
{
  ONE = 1,
  TWO,
  THREE,
  FIVE = 5,
  SIX,
  EIGHT = 8
}
```

The generated Go code typedefs Numberz to int32 and can deserialize from any int32 value.
```go
type Numberz int32

func (v *Numberz) FromWire(w wire.Value) error {
	*v = (Numberz)(w.GetI32())
	return nil
}
```

The generated Java code uses an enum type with `null` as a fallback.
```java
  public static Numberz findByValue(int value) { 
    switch (value) {
      case 1:
        return ONE;
      case 2:
        return TWO;
      case 3:
        return THREE;
      case 5:
        return FIVE;
      case 6:
        return SIX;
      case 8:
        return EIGHT;
      default:
        return null;
    }
  }
```

The Go crossdock client is expecting `Numberz(42)` to be valid, but this is not deserializable in Java.  The Java code appears to be doing the right thing, so I believe this is a bug in thriftrw? @abhinav

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/gauntlet/ThriftTestHandler.java
```diff
@@ -0,0 +1,260 @@
...
+
+  @Override
+  public CompletionStage<Reply<Void>> testVoid(Context ctx) {
+    return CompletableFuture.completedFuture(DefaultReply.<Void>builder().build());
+  }
+
+  @Override
+  public CompletionStage<Reply<String>> testString(Context ctx, String thing) {
+    return CompletableFuture.completedFuture(
+        DefaultReply.<String>builder().headers(new Headers(ctx.getHeaders())).body(thing).build());
```
bpon: Are headers supposed to be echoed?

vivekrsharma:
Yes, both the test failures are due to violation of the enum contract by Numberz. All the other tests are passing otherwise.

PR #325, opened 2017-06-05
[Part 1]: Grpc crossdock testing support

vivekrsharma:
Adds support for grpc testing  and various encodings to ensure we've some regression tests. This part supports proto over http and tests it against go server. Test support for proto over grpc will be followed in later PRs.


breerly:
Just to be super clear - this adds testing of *pure* grpc-java w/ yarpc-go?

vivekrsharma:
Yes @breerly. working on polishing stuff but mostly the structure is going to look similar.

config/findbugs/exclude.xml
```diff
@@ -4,6 +4,7 @@
     <Or>
       <Class name="~.*\.generated\..*"/>
       <Class name="~thrift\.test\..*"/>
+      <Class name="~.*/generated-sources/..*"/>
```
peter-edge: Should this be `generated` or `generated-sources` re: change above?

docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
             - AXIS_CLIENT=go,java,node,python,python-sync
             - AXIS_SERVER=go,java,node,python
             - AXIS_TRANSPORT=http
-            - AXIS_ENCODING=json,raw,thrift
+            - AXIS_ENCODING=protobuf #json,raw,thrift,protobuf
```
peter-edge: ?

docker-compose.yml
```diff
@@ -33,6 +33,8 @@ services:
             # Transports available to the ctxpropagation behavior for multihop requests.
             - AXIS_CTXAVAILABLETRANSPORTS=http
 
+            - BEHAVIOR_PROTOBUF=client,server,transport
```
peter-edge: Help me understand why this is all in yarpc-java, do tests run from here as well?

yarpc-crossdock/build.gradle
```diff
@@ -1,16 +1,37 @@
 import java.nio.file.Paths
 
 apply plugin: 'org.jruyi.thrift'
+apply plugin: 'java'
```
peter-edge: Plugin `java`?

config/checkstyle/suppressions.xml
```diff
@@ -6,6 +6,8 @@
 <suppressions>
     <!-- Ignore files inside a gen-java directory. -->
     <suppress files="[\\/]generated-sources[\\/]" checks="."/>
+    <!-- Ignore files inside a proto directory. -->
+    <suppress files="[\\/]generated[\\/]" checks="."/>
```
peter-edge: It seems from the gradle file that this is `generated-sources`? Also see comment on `config/findbugs/exclude.xml`.

yarpc-crossdock/build.gradle
```diff
@@ -1,16 +1,37 @@
...
     compile project(':yarpc-core')
     compile project(':yarpc-encoding-thrift')
     compile project(':yarpc-transport-http')
     compile project(':yarpc-encoding-raw')
     compile project(':yarpc-encoding-json')
 
+    compile 'io.grpc:grpc-netty:1.3.0'
+    compile 'io.grpc:grpc-protobuf:1.3.0'
+    compile 'io.grpc:grpc-stub:1.3.0'
+    compile 'com.google.protobuf:protobuf-java:3.0.0'
```
peter-edge: This is `3.3.0` for main protobuf, is this different for java?

yarpc-crossdock/build.gradle
```diff
@@ -30,6 +51,43 @@ compileThrift {
...
+  protoc {
+    // The artifact spec for the Protobuf Compiler
+    artifact = 'com.google.protobuf:protoc:3.0.0'
+  }
+  plugins {
+    // Optional: an artifact spec for a protoc plugin, with "grpc" as
+    // the identifier, which can be referred to in the "plugins"
+    // container of the "generateProtoTasks" closure.
+    grpc {
+      artifact = 'io.grpc:protoc-gen-grpc-java:1.0.0-pre2'
```
peter-edge: Is this the right version?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/proto/EchoServer.java
```diff
@@ -0,0 +1,65 @@
...
+
+@Slf4j
+public class EchoServer {
+
+  private Server server;
+
+  /** Starts the server. */
+  public void start() throws IOException {
+    /* The port on which the server should run */
+    int port = 8090;
```
peter-edge: `8089`

yarpc-crossdock/src/main/java/io/yarpc/crossdock/behavior/proto/EchoServer.java
```diff
@@ -0,0 +1,65 @@
...
+    server = ServerBuilder.forPort(port).addService(new EchoProtoHandler()).build().start();
+    log.info("Server started, listening on " + port);
+    Runtime.getRuntime()
+        .addShutdownHook(
+            new Thread() {
+              @Override
+              public void run() {
+                System.err.println("*** shutting down gRPC server since JVM is shutting down");
```
peter-edge: Is this right for logging?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/proto/EchoClient.java
```diff
@@ -0,0 +1,77 @@
...
+    blockingStub = EchoGrpc.newBlockingStub(channel);
+  }
+
+  public void shutdown() throws InterruptedException {
+    channel.shutdown().awaitTermination(5, TimeUnit.SECONDS);
+  }
+
+  /** Echo to server. */
+  public String echo(String token) {
+    log.info("Will try to echo");
```
peter-edge: Do you need all these log statements? They could pollute some debugging, idk, not sure

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/proto/EchoProtoBehavior.java
```diff
@@ -0,0 +1,47 @@
...
+import works.crossdock.CrossdockResponse;
+import works.crossdock.client.Behavior;
+
+@Slf4j
+public class EchoProtoBehavior implements Behavior {
+  @SuppressFBWarnings("NP_NONNULL_PARAM_VIOLATION")
+  @Override
+  public CompletionStage<CrossdockResponse> run(CrossdockRequest request) {
+    log.info("Crossdock request received");
+    EchoClient client = new EchoClient("localhost", 8090);
```
peter-edge: `8089`

yarpc-crossdock/src/main/proto/yarpc.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package uber.yarpc;
+
+option java_package = "io.yarpc.proto.generated";
```
peter-edge: `com.uber.yarpc`? Also this should be in yarpc-go, but we can deal with that later

config/checkstyle/suppressions.xml
vivekrsharma: It is generated-sources. The gradle plugin generates sources under generated-sources. Would that be a problem?

config/findbugs/exclude.xml
vivekrsharma: Commented above.

docker-compose.yml
vivekrsharma: good point. Yes, even java is going to launch these tests. Our last diff will enable this even in yarpc-go.

yarpc-crossdock/build.gradle
vivekrsharma: addressed, don't need this.

yarpc-crossdock/build.gradle
vivekrsharma: Let's keep it same everywhere. Changed it to 3.3.0

yarpc-crossdock/build.gradle
vivekrsharma: so I followed the exampleProject published by protobuf-gradle-plugin here. https://github.com/google/protobuf-gradle-plugin/blob/master/exampleProject/build.gradle

It seemed to me like this is maintained by Google and hence I believe this combination would be right to go with. 

However we will maintain uniformity across go/java and change the com.google.protofbuf:protoc:3.3.0

yarpc-crossdock/src/main/proto/yarpc.proto
vivekrsharma: ummm, I tried to stay consistent with thrift. Thrift gets generated under io.yarpc.thrift.generated

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/proto/EchoClient.java
vivekrsharma: nope, this is garbage. Sorry this escaped.

docker-compose.yml
peter-edge: Is this just a copy from yarpc-go?

docker-compose.yml
```diff
@@ -64,14 +68,17 @@ services:
         # TODO: change the image to yarpc/yarpc-go after release that
         # fixes https://github.com/yarpc/yarpc-go/issues/777
         image: yarpc/yarpc-go:dev
+#        build:
+#          context: /Users/viveksharma/gocode/src/go.uber.org/yarpc
```
peter-edge: ?

docker-compose.yml
vivekrsharma: yes.

yarpc-crossdock/build.gradle
```diff
@@ -1,16 +1,36 @@
 import java.nio.file.Paths
 
 apply plugin: 'org.jruyi.thrift'
+apply plugin: 'com.google.protobuf'
 
 description = 'Crossdock test library for yarpc-java'
 
+repositories {
```
bpon: I'm pretty sure you don't need this section.

yarpc-crossdock/build.gradle
```diff
@@ -1,16 +1,36 @@
...
+  maven { url "https://plugins.gradle.org/m2/" }
+}
+
+buildscript {
```
bpon: This will work, but it's the old way of discovering plugins.  You can actually get rid of this entire block and just add the following to the `plugins` section of the root `build.gradle`:

```
id "com.google.protobuf" version "0.8.1" apply false
```

Now that I'm looking at the root `build.gradle`, some of those other plugins should also have `apply false`, which basically means that you're declaring that you'll use the plugin somewhere, but it doesn't need to be applied right away (because you'll explicitly apply it to subprojects that need it).

yarpc-crossdock/build.gradle
bpon: The latest version is 1.3.0.

https://mvnrepository.com/artifact/io.grpc/protoc-gen-grpc-java

yarpc-crossdock/build.gradle
```diff
@@ -30,6 +50,43 @@ compileThrift {
...
+
+clean {
+    delete protobuf.generatedFilesBaseDir
+}
+
+idea {
+    module {
+        sourceDirs += file("${protobuf.generatedFilesBaseDir}/main/java");
+        // If you have additional sourceSets and/or codegen plugins, add all of them
+        sourceDirs += file("${protobuf.generatedFilesBaseDir}/main/grpc");
```
bpon: Does this directory even exist?

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/proto/EchoClient.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /** Echo to server. */
+  public String echo(String token) {
+    Ping request = Ping.newBuilder().setBeep(token).build();
+    Pong response;
+    try {
+      response = blockingStub.echo(request);
+    } catch (StatusRuntimeException e) {
+      log.warn("RPC failed: {0}", e.getStatus());
```
bpon: I don't think slf4j supports indexed arguments.  I've only ever seen "{}".

yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/proto/EchoProtoBehavior.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import lombok.extern.slf4j.Slf4j;
+import works.crossdock.CrossdockRequest;
+import works.crossdock.CrossdockResponse;
+import works.crossdock.client.Behavior;
+
+@Slf4j
+public class EchoProtoBehavior implements Behavior {
+  @SuppressFBWarnings("NP_NONNULL_PARAM_VIOLATION")
```
bpon: This looks unnecessary.

yarpc-crossdock/src/main/proto/yarpc.proto
bpon: There's no need to be consistent with thrift.  If there is a `package` declaration, we should honor it as much as possible (with perhaps some minor tweaks for Java conventions).  Like Peter suggests, `com.uber.yarpc` might be appropriate.  I'd also be ok with just removing this line and just letting it default to `package uber.yarpc`.  In any case, it doesn't matter too much since it's just used for crossdock.

PR #326, opened 2017-06-13
[Part2] grpc: add yarpc headers

vivekrsharma:
Adds yarpc headers and gets the server from the crossdock request.


yarpc-crossdock/src/main/java/io/yarpc/crossdock/client/proto/EchoProtoBehavior.java
```diff
@@ -32,7 +32,8 @@
 public class EchoProtoBehavior implements Behavior {
   @Override
   public CompletionStage<CrossdockResponse> run(CrossdockRequest request) {
-    EchoClient client = new EchoClient("localhost", 8089);
+    log.info("Inside run for proto behavior");
```
peter-edge: is this needed?

vivekrsharma: Deleted.

